PR #12, opened 2023-05-02
Implementation of Mahalanobis classifier and transformer

grovduck:
- Add tests from yaImpute Moscow mahalanobis run (k=5)
- Only pass X on call to MyStandardScaler in Euclidean and Mahalanobis

More or less verbatim port from `yaImpute` using `numpy` functions.


grovduck:
@aazuspan, this is a trivial port, but a few things to consider:

- Right now, I'm mixing names on classes (`Euclidean` and `Mahalanobis`) with tests (`euc` and `mah`).  Better to be consistent?  I'd want to change the tests to be the full names.
- Is this the right time to change the `original` module to `transformers`?  We'll have an `MSN` transformer in there as well soon.
- It might be too early for this (or not the right PR), but `predict` in each of `Euclidean`, `Mahalanobis`, and `GNN` are identical and should probably go into `TransformedKNeighborsMixin`, right?

src/sklearn_knn/original/_mahalanobis_transformer.py
```diff
@@ -0,0 +1,18 @@
+import numpy as np
+from sklearn.base import BaseEstimator, TransformerMixin
+
+from .._base import MyStandardScaler
+
+
+class MahalanobisTransformer(TransformerMixin, BaseEstimator):
+    def fit(self, X, y=None):
+        self.scalar_ = MyStandardScaler().fit(X)
```
aazuspan, line 9: Does it make more sense to call this a `scaler_` than a `scalar_`? Scalar makes me think of the mathematical or programming meaning, which I don't think apply here?

aazuspan:
Awesome, this looks great! I won't pretend to understand how it's working, but if it passes the port tests I'm happy. I just had one nitpick on naming.

> Better to be consistent? I'd want to change the tests to be the full names.

Yeah, I think I agree. We're not saving anything by using shorter names, so might as well keep it consistent with the full names.

> Is this the right time to change the original module to transformers?

I don't see any harm in doing that here. I'd say go for it.

> It might be too early for this (or not the right PR), but predict in each of Euclidean, Mahalanobis, and GNN are identical and should probably go into TransformedKNeighborsMixin, right?

Good point! There's still a lot of work that needs to go into the `predict` method to get it to actually run imputing and handle multiple neighbors with weighting, etc, right? Do you have a clear idea of whether that implementation will be the same for all the estimators? If so, I wouldn't object to consolidating it now.

src/sklearn_knn/original/_mahalanobis_transformer.py
grovduck: Yep, of course you're right

grovduck:
>  I won't pretend to understand how it's working

Dirty little secret, I don't really either - I'm mostly porting.

> I just had one nitpick on naming.

Nitpicking is always encouraged!

> Yeah, I think I agree. We're not saving anything by using shorter names, so might as well keep it consistent with the full names.

Cool, changed.

> I don't see any harm in doing that here. I'd say go for it.

Changed as well.

> Good point! There's still a lot of work that needs to go into the predict method to get it to actually run imputing and handle multiple neighbors with weighting, etc, right? Do you have a clear idea of whether that implementation will be the same for all the estimators? If so, I wouldn't object to consolidating it now.

My thought is that the four methods that share this (`Euclidean`, `Mahalanobis`, `MSN`, and `GNN`) should all be capable of handling weighting and multiple neighbors in the same way (since they are effectively being transformed into a Euclidean space before before neighbor finding).  The only possible exception is that `Euclidean` and `Mahalanobis` shouldn't subset their axes as it's one "axis" per covariate, whereas in `MSN` and `GNN`, each axis is a linear combination of all covariates.  This gives me enough pause to not tackle this in this PR and wait until we have `MSN` in place and start really thinking through how we're going to handle `predict` with multiple atributes.  Thanks for the encouragement to reflect!

aazuspan (approved):

aazuspan:
Great, I'd say we're good to merge!

grovduck:
Sweet, my first merge!  So, would you do a merge commit or a squash and merge here?  Only two commits with the last being pretty inconsequential.  Do you have any guidelines/advice on when to do each?

aazuspan:
First of many! I would probably squash merge, but I can't say that's based on any consistent criteria. Probably the deciding factor should be whether a squash merge would erase any valuable commit history, which as you mentioned wouldn't really be the case here.

Issue #2, opened 2023-04-14
Decide how to handle fitting, prediction, and ID data


aazuspan:
See discussion in #1.

Some estimators can be fit with different data than they will predict (e.g. fitting with species composition data and predicting structural attributes using GNN), so we need to be able to set that data (tentatively called `y_predict`) somehow. We also need to the flexibility to set different predicted data without requiring refitting.

A few options discussed:

1. Accept `y_predict` as a parameter of `fit` to store for later use. If this is the only way to pass `y_predict`, it would require refitting when switching predicted attributes, so this alone is not a complete solution.
2. Use a `set_target_data` method on estimators after fitting. This gives the flexibility of being able to switch data without refitting, but would interrupt the normal `fit` then `predict` then `score` workflow.
3. Accept `y_predict` as a parameter of `predict`. Need to think this through further.

grovduck:
Hey @aazuspan, I've been thinking about this and how it relates to #3.  My thinking is still a bit muddy, but here is where I am currently.

In order for `predict` and `score` to work somewhat seamlessly for these estimators *without* fundamentally changing the signature for those methods, I keep coming back to the thought that `y` matrix passed to `fit` should include the attributes we want to predict (what we've called `y_predict` above) and not passing it as a separate keyword argument of `fit`.  Looking at the `linnerud` multi-output data has changed my mind a bit on this.  Here's a toy example mimicking the format of what the `linnerud` data would give you:
```
# Assume the y are the attributes we want to predict, not the same as the species matrix
>>> from sklearn.neighbors import KNeighborsClassifier, KNeighborsRegressor
>>> X = [[1, 2, 3], [1, 2, 4], [2, 2, 3]]   # like load_linnerud().data
>>> y = [[5, 10, 15], [3, 6, 9], [4, 8, 12]]  # like load_linnerud().target
>>> X_trg = [[2, 2, 3.5]]

# With KNeighborsClassifier, this is the same as k=1
>>> clf = KNeighborsClassifier(n_neighbors=2, weights="distance").fit(X, y)
>>> clf.kneighbors(X_trg)
(array([[0.5       , 1.11803399]]), array([[2, 1]]))
>>> clf.predict(X_trg)
[[ 4  8 12]]

# With KNeighborsRegressor, this is the same as k > 1 (in this case using inverse distance as weights)
>>> clf = KNeighborsRegressor(n_neighbors=2, weights="distance").fit(X, y)
[...]

PR #14, opened 2023-05-08
Subclass estimators from KNeighborsRegressor

### LINKED ISSUE #13: Support continuous multioutput targets
This ties in to #2, but I think is going to need its own issue to track: now that we're planning on passing the target data (e.g. composition) directly in as `y`, subclassing from `KNeighborsClassifier` no longer works because it doesn't support continuous multioutput targets. For example:

```python
clf = Mahalanobis()
X = [[0, 1], [1, 1], [2, 2], [3, 3]]  # environmental features
y = [[0.1, 0.2], [0.2, 0.3], [0.3, 0.4], [0.4, 0.5]]  # composition targets
clf.fit(X, y)

ValueError: Unknown label type: 'continuous-multioutput'
```

The easiest solution is to subclass `KNeighborsRegressor` instead, which supports continuous multioutput targets. This will break the current `kneighbor_ids` implementation which relies on the `classes_` attribute that doesn't exist on regressors, but I think that was going to need to be modified regardless based our discussion on handling plot IDs. There might be other side effects to this change I haven't considered...

The other option would be to patch continuous multioutput support into a subclass of `KNeighborsClassifier`, and subclass the rest of our estimators from that instead. This sort of gets back to #6 and the question of whether we're going to consider our estimators as regressors, classifiers, or something in between.

Let me know what you think!

aazuspan:
This would close #13 by making all estimators subclass `KNeighborsRegressor` instead of `KNeighborsClassifier`, allowing us to fit and predict continuous multioutput data. It also removes the `kneighbor_ids` method following the discussion in #2, but leaves the (renamed) `IDNeighborsRegressor` in place while we figure out whether we need that. 

All of the tests against neighbor IDs have been temporarily disabled. If we end up supporting dataframe indices, we'll probably want to modify to test for those. If not, we can probably get rid of them unless there's a reason to test array indices (which I assume would require remaking the neighbor csvs).

Lastly, this adds some very basic tests to make sure all the estimators can `fit` and `predict` with continuous multioutput data. I'm assuming those will be redundant with the `sklearn.utils.estimator_checks`, so we may remove them at that time or alternatively expand the tests to make sure the predictions are accurate (in which case they'd probably move into the port tests).

### DIFF
--- a/src/sklearn_knn/_base.py
+++ b/src/sklearn_knn/_base.py
@@ -1,24 +1,19 @@
 import numpy as np
-from sklearn.neighbors import KNeighborsClassifier
+from sklearn.neighbors import KNeighborsRegressor
 from sklearn.preprocessing import StandardScaler
 from sklearn.utils.validation import check_is_fitted
 
 
-class IDNeighborsClassifier(KNeighborsClassifier):
+class IDNeighborsRegressor(KNeighborsRegressor):
     """
-    Specialized KNeighborsClassifier where labels
-    are IDs for samples and not classes
+    Placeholder class for implementing plot ID access.
     """
 
-    def kneighbor_ids(self, X=None, n_neighbors=None):
-        neigh_ind = super().kneighbors(X, n_neighbors, False)
-        return self.classes_[self._y[neigh_ind]]
 
-
-class TransformedKNeighborsMixin(KNeighborsClassifier):
+class TransformedKNeighborsMixin(KNeighborsRegressor):
     """
-    Mixin for KNeighbors classifiers that store a `transform_` during fitting
-    (e.g. GNN).
+    Mixin for KNeighbors regressors that store a `transform_` during fitting (e.g.
+    GNN).
     """
 
     def kneighbors(self, X=None, n_neighbors=None, return_distance=True):
--- a/src/sklearn_knn/_euclidean.py
+++ b/src/sklearn_knn/_euclidean.py
@@ -1,9 +1,9 @@
 from sklearn.utils.validation import check_is_fitted
 
-from ._base import IDNeighborsClassifier, MyStandardScaler, TransformedKNeighborsMixin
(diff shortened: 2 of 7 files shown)


grovduck:
@aazuspan, this looks great to me.  I didn't see any necessary changes.

aazuspan:
Great, I'll merge it!

grovduck:
Sorry to be commenting on closed pull requests, but had a thought on one of your comments:

> All of the tests against neighbor IDs have been temporarily disabled. If we end up supporting dataframe indices, we'll probably want to modify to test for those. If not, we can probably get rid of them unless there's a reason to test array indices (which I assume would require remaking the neighbor csvs).

If we do want to test indexes, I think this should work - creating a dictionary between plot ID and index and then using `df.applymap` to spread to all other columns.

conftest.py
```
def set_lookup_dict(df):
    return dict(zip(df.ID, df.index))


def id_to_idx(df, lookup_dict):
    def lookup(id_):
        return lookup_dict.get(id_, id_)

    return df.applymap(lookup)

class Dataset:
    def __init__(self, project, method, k=5):
        ...
        lookup_dict = set_lookup_dict(ref_neighbors_df)
        self.ref_idxs = id_to_idx(ref_neighbors_df, lookup_dict).loc[:, cols].values
        self.trg_idxs = id_to_idx(trg_neighbors_df, lookup_dict).loc[:, cols].values
```
test_port.py
```
def test_moscow_raw(moscow_raw):
    ...
    assert_array_equal(nn, moscow_raw.ref_idxs)
    assert_array_equal(nn, moscow_raw.trg_idxs)
```
I don't know if you think it's a valuable test or not, but it should avoid recreating the test CSV files

aazuspan:
Thanks for putting this together! I haven't tried running it, but this looks like a nice, clever solution if we end up going this route.

PR #18, opened 2023-05-10
Rename estimators and package

### LINKED ISSUE #6: Finalize naming conventions for estimators and package
Currently, our estimators are named to follow method names introduced in R's `yaImpute` package.  We likely want to conform to scikit-learn's model of being more explicit about whether estimators are classifiers or regressors.  kNN methods fit into a funny space here - most often they are thought of as classifiers as they use the `k` nearest neighbors in prediction.  However, some attributes are continuous and are calculated as (weighted) means from the `k` neighbors attributes, which is more similar to [KNeighborsRegressor](https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.KNeighborsRegressor.html).

Also need to decide whether it makes sense to stick with `scikit-learn-knn` as a package name as `knn` is already a meaningful term in `scikit-learn`.  We could also decide to "pay homage" to `yaImpute` which is serving as the inspiration for this package.

grovduck:
This PR would close #6 by renaming all current estimators to specify that they are `KnnRegressor` estimators.  It also changing the package name to `scikit-learn-knn-regression`, although we haven't yet renamed this in Github, but will before we merge this change.  The imported module name also changes from `sklearn_knn` to `sknnr`.

To make the change locally, I did these steps:

1. renamed all instances of `scikit-learn-knn` to `scikit-learn-knn-regression` in all files
2. renamed all instances of `sklearn_knn` to `sknnr` in all files
3. renamed estimators to be suffixed with `KnnRegressor` (`Raw`, `Euclidean`, `Mahalanobis`) or `Regressor` (`GNN`, `MSN`)
4. changed directory name `src\sklearn_knn` to `src\sknnr`
5. `hatch env prune` to get rid of `scikit-learn-knn` and `test` environments
6. `hatch shell` to add `scikit-learn-knn-regression` environment
7. `hatch run test:all` to add `test` environment

(The only hiccup I had was that the old `scikit-learn-knn` hatch environments were not deleted, but I can do that manually).


grovduck:
@aazuspan , a couple of possible changes to consider:

1. Do we want e.g. `EuclideanKnnRegressor` or `EuclideanKNNRegressor`?  I'm OK either way.
2. Doing a bit of browsing on PyPI, it looks like both `scikit-learn` and `sklearn` prefixes are used for custom estimators.  `scikit-learn` is more explicit, `sklearn` is shorter for an already long package name.

README.md
aazuspan, line 1: Can you update the heading on line 1? Thanks!

aazuspan (changes requested):
Looks great overall!

> Do we want e.g. EuclideanKnnRegressor or EuclideanKNNRegressor? I'm OK either way.

I lean towards `KNN` just for consistency with `GNN` and `MSN`, but not a strong opinion.

> Doing a bit of browsing on PyPI, it looks like both scikit-learn and sklearn prefixes are used for custom estimators. scikit-learn is more explicit, sklearn is shorter for an already long package name.

Good points. I guess if I had to choose I would go with the more explicit `scikit-learn`, but if you have any preference I'd happily defer to that.

grovduck:
> I lean towards `KNN` just for consistency with `GNN` and `MSN`

I agree.  Changed to `KNN` in this commit.

> I would go with the more explicit `scikit-learn`

I don't have a strong opinion, but agree that explicit is likely better.

aazuspan (approved):
Perfect, looks good to merge!

grovduck:
Cool, I've renamed the repository to `scikit-learn-knn-regression` on Github and run this on my local repo:

`git remote set-url origin git@github.com:lemma-osu/scikit-learn-knn-regression.git`

aazuspan:
Awesome, thanks!

Issue #21, opened 2023-05-13
Get all `sklearn` estimator checks passing


aazuspan:
As discussed in #20, we've run into a number of issues that probably would have been caught earlier with the built-in test suite in the `sklearn.utils.estimator_checks` module, so it's probably in our best interest to focus on getting those passing. This should also potentially allow us to remove some redundant tests from `test_estimators.py`.

Note that it seems like some valuable tests like [check_dataframe_column_names_consistency](https://github.com/scikit-learn/scikit-learn/blob/72a604975102b2d93082385d7a5a7033886cc825/sklearn/utils/estimator_checks.py#L3873) aren't running automatically, so we'll need to make sure we run everything that's relevant.

EDIT: Our tests currently contain commented-out checks for our estimators, but it looks like there are also [transformer checks](https://github.com/scikit-learn/scikit-learn/blob/99c9f41c774b6b3d4f36f515b06f5b20838d0771/sklearn/utils/estimator_checks.py#LL324C15-L324C15) that we should run. In fact, maybe it would make sense to get our transformers passing first since those will affect the estimator results?

Another aspect that I'm not very familiar with but that we'll probably need to dive into for this is the [sklearn tag system](https://scikit-learn.org/stable/developers/develop.html#estimator-tags), which allows you to specify different compatibility options for estimators and transformers that inform which checks will be run.

PR #22, opened 2023-05-15
Store feature names for transformed estimators

### LINKED ISSUE #20: Transformed regressors drop dataframe feature names
Hey @grovduck, I'm starting to sound like a broken record, but there's *another* issue blocking dataframe indexes in #2. This is a bit of a long one, so I tried to lay it out below.

## The problem

All of our `TransformedKNeighborsMixin` estimators are incompatible with dataframes in that they don't store feature names. I didn't think to check this before, but updating the dataframe test to check for feature names fails for everything but `RawKNNRegressor`.

```python
@pytest.mark.parametrize("estimator", get_kneighbor_estimator_instances())
def test_estimators_support_dataframes(estimator, moscow_euclidean):
    """All estimators should fit and predict data stored as dataframes."""
    num_features = moscow_euclidean.X.shape[1]
    feature_names = [f"col_{i}" for i in range(num_features)]

    X_df = pd.DataFrame(moscow_euclidean.X, columns=feature_names)
    y_df = pd.DataFrame(moscow_euclidean.y)

    estimator.fit(X_df, y_df)
    estimator.predict(X_df)
    assert_array_equal(estimator.feature_names_in_, feature_names)
```

This happens because they all run `X` through transformers that convert the dataframes to arrays before they get to `KNeighborsRegressor.fit` where the features would be retrieved and stored. The same thing would happen with `sklearn` transformers, so I think we should probably solve this in `TransformedKNeighborsMixin` rather than in the transformers.

EDIT: As detailed in the next post, once we solve the issue of losing feature names when fitting, we need to also retain feature names when predicting to avoid warnings.

## Possible solutions

First of all, I think we should move the [actual transformation](https://github.com/lemma-osu/scikit-learn-knn-regression/blob/8b27404632e79af9f3b37b41006692fc76e47db8/src/sknnr/_euclidean.py#L10) out of the `fit` method for each estimator and into a `fit` method for `TransformedKNeighborsMixin`. That should probably be done regardless of this issue just to reduce some duplication, and also allows us to make sure everything gets fit the same way. Then, I think we need to modify that `fit` method to make sure it sets appropriate feature names after transformation.

To get feature names, [all that sklearn does](https://github.com/scikit-learn/scikit-learn/blob/2a2772a87b6c772dc3b8292bcffb990ce27515a8/sklearn/utils/validation.py#L1870) is look for a `columns` attribute on `X`. If we could copy that `columns` attribute onto the transformed array before passing it to `KNeighborsRegressor.fit` we'd be set, but there's [no way to directly set attributes on Numpy arrays](https://stackoverflow.com/questions/67509913/add-an-attribute-to-a-numpy-array-in-runtime) because they are implemented in C.

I think that leaves us with a few options:

1. Use [`sklearn.utils.validation._get_feature_names`](https://github.com/scikit-learn/scikit-learn/blob/2a2772a87b6c772dc3b8292bcffb990ce27515a8/sklearn/utils/validation.py#L1870) to get and validate the feature names before transforming, then manually set them as `feature_names_in_` after fitting. I don't love this because it requires us to use a private method that could disappear, get renamed, change return types, etc. The upside is that we would know our feature names are retrieved consistently with `sklearn`. 
2. Same as above but copy `sklearn.utils.validation._get_feature_names` into our code base. That bypasses the private method issue, but adds some maintenance cost and we would need to carefully consider how to do that consistently with the `sklearn` license. As with option 1, we would still need to handle setting the `feature_names_in_` attribute.
3. [Subclass `ndarray`](https://numpy.org/doc/stable/user/basics.subclassing.html#simple-example-adding-an-extra-attribute-to-ndarray) to support a `column` attribute and pass that in to fit. As long as `sklearn` doesn't change how they identify features (which seems unlikely), we could let `sklearn` handle getting and setting feature names, and I *think* it would be transparent to users. I did confirm that the `_fit_X` attribute seems to store a numpy array regardless of what goes into it. Like option 2, this adds some maintenance cost.
4. Use the `_check_feature_names` method with the non-transformed `X` after fitting. This will set feature names on the model and fix the issue of losing feature names when fitting. The downside is that we're again using a private method.

I don't love any of these options, so let me know what you think or if any other solutions occur to you.

## Estimator checks

I noticed that the `sklearn.estimator_checks` [would have caught this](https://github.com/scikit-learn/scikit-learn/blob/72a604975102b2d93082385d7a5a7033886cc825/sklearn/utils/estimator_checks.py#L3918), so I wonder if we should prioritize getting those checks to pass before we add any more functionality? I think that may be a big lift, but would at least prevent us from accidentally breaking estimators in the future. Also, it may be easier to do now than after they get more complex and would keep us from accidentally writing a lot of redundant tests.

EDIT: This would also catch warnings for predicting without feature names that I mention in the next post.

aazuspan:
Hey @grovduck, I decided to move ahead with this since, as you mentioned, it should hopefully simplify #21 a little bit.

This would close #20 by wrapping dataframe `X` arrays in the new `NamedFeatureArray` after transforming and before passing them on `fit`, `predict`, or `kneighbors` methods. By storing their `columns` attribute, this allows `sklearn` to access and set feature names that would otherwise be lost.

A few questions/things for you to consider as you look over this:

1. What do you think about having `NamedFeatureArray` in `_base`? It's not directly related to `sklearn` which makes me think it could go elsewhere, but let me know what you think.
2. This moves `predict` out of the individual estimators and into `TransformedKNeighborsMixin`. When we discussed this before we decided to leave `predict` duplicated in our estimators for now in case they need to be implemented differently, but since this would require modifying them all anyways, I went ahead and combined them. Let me know if you think that's premature and I should re-implement them in the subclasses.
3. To keep things simple, I just wrote a manual test for `feature_names_in_` rather than trying to take advantage of the `estimator_checks` module. I figure we'll have to get a lot more familiar with that module as we work on #21, so if it makes sense we can switch to one of the built-in tests then.

Just noticed there's definitely a typo in the commit title 🤦‍♂️

### DIFF
--- a/src/sknnr/_base.py
+++ b/src/sknnr/_base.py
@@ -1,5 +1,24 @@
+import numpy as np
 from sklearn.neighbors import KNeighborsRegressor
 from sklearn.utils.validation import check_is_fitted
+
+
+class NamedFeatureArray(np.ndarray):
+    """An array with a columns attribute indicating feature names.
+
+    Storing a `columns` attribute allows this array to act like  a dataframe for the
+    purpose of extracting feature names when passed to sklearn estimators.
+    """
+
+    def __new__(cls, array, columns=None):
+        obj = np.asarray(array).view(cls)
+        obj.columns = columns
+        return obj
+
+    def __array_finalize__(self, obj):
+        if obj is None:
+            return
+        self.columns = getattr(obj, "columns", None)
 
 
 class IDNeighborsRegressor(KNeighborsRegressor):
@@ -10,14 +29,39 @@ class IDNeighborsRegressor(KNeighborsRegressor):
 
 class TransformedKNeighborsMixin(KNeighborsRegressor):
     """
-    Mixin for KNeighbors regressors that store a `transform_` during fitting (e.g.
-    GNN).
+    Mixin for KNeighbors regressors that apply transformations to the feature data.
     """
 
+    def _transform(self, X) -> NamedFeatureArray:
+        """Apply the stored transform to the input data.
+
+        Note
+        ----
+        Transforming will cast input data to numpy arrays. To preserve feature names
+        in the case of dataframe inputs, this method will wrap the transformed array
+        in a `NamedFeatureArray` with a `columns` attribute, allowing `sklearn` to
+        parse and store feature names.
+        """
+        check_is_fitted(self, "transform_")
+        X_transformed = self.transform_.transform(X)
+        if hasattr(X, "columns"):
+            X_transformed = NamedFeatureArray(X_transformed, columns=X.columns)
+
+        return X_transformed
+
+    def fit(self, X, y):
+        """Fit using transformed feature data."""
+        X_transformed = self._transform(X)
+        return super().fit(X_transformed, y)
+
+    def predict(self, X):
+        """Predict using transformed feature data."""
+        X_transformed = self._transform(X)
+        return super().predict(X_transformed)
+
     def kneighbors(self, X=None, n_neighbors=None, return_distance=True):
-        if X is not None:
-            check_is_fitted(self, "transform_")
-            X = self.transform_.transform(X)
+        """Return neighbor indices and distances using transformed feature data."""
+        X_transformed = self._transform(X) if X is not None else X
         return super().kneighbors(
-            X=X, n_neighbors=n_neighbors, return_distance=return_distance
+            X=X_transformed, n_neighbors=n_neighbors, return_distance=return_distance
         )
--- a/src/sknnr/_euclidean.py
+++ b/src/sknnr/_euclidean.py
@@ -1,5 +1,3 @@
-from sklearn.utils.validation import check_is_fitted
-
 from ._base import IDNeighborsRegressor, TransformedKNeighborsMixin
 from .transformers import StandardScalerWithDOF
 
@@ -7,10 +5,4 @@ from .transformers import StandardScalerWithDOF
 class EuclideanKNNRegressor(IDNeighborsRegressor, TransformedKNeighborsMixin):
     def fit(self, X, y):
         self.transform_ = StandardScalerWithDOF(ddof=1).fit(X)
-        X = self.transform_.transform(X)
         return super().fit(X, y)
-
-    def predict(self, X):
-        check_is_fitted(self)
-        X = self.transform_.transform(X)
-        return super().predict(X)
--- a/src/sknnr/_gnn.py
+++ b/src/sknnr/_gnn.py
@@ -1,5 +1,3 @@
-from sklearn.utils.validation import check_is_fitted
-
 from ._base import IDNeighborsRegressor, TransformedKNeighborsMixin
 from .transformers import CCATransformer
 
@@ -7,10 +5,4 @@ from .transformers import CCATransformer
 class GNNRegressor(IDNeighborsRegressor, TransformedKNeighborsMixin):
     def fit(self, X, y, spp=None):
         self.transform_ = CCATransformer().fit(X, y=y, spp=spp)
-        X = self.transform_.transform(X)
         return super().fit(X, y)
-
-    def predict(self, X):
-        check_is_fitted(self)
-        X = self.transform_.transform(X)
-        return super().predict(X)
--- a/src/sknnr/_mahalanobis.py
+++ b/src/sknnr/_mahalanobis.py
@@ -1,5 +1,3 @@
-from sklearn.utils.validation import check_is_fitted
-
 from ._base import IDNeighborsRegressor, TransformedKNeighborsMixin
 from .transformers import MahalanobisTransformer
 
@@ -7,10 +5,4 @@ from .transformers import MahalanobisTransformer
 class MahalanobisKNNRegressor(IDNeighborsRegressor, TransformedKNeighborsMixin):
     def fit(self, X, y):
         self.transform_ = MahalanobisTransformer().fit(X)
-        X = self.transform_.transform(X)
         return super().fit(X, y)
-
-    def predict(self, X):
-        check_is_fitted(self)
-        X = self.transform_.transform(X)
-        return super().predict(X)
--- a/tests/test_estimators.py
+++ b/tests/test_estimators.py
@@ -2,6 +2,7 @@ from typing import List
 
 import pandas as pd
 import pytest
+from numpy.testing import assert_array_equal
 
 # from sklearn.utils.estimator_checks import parametrize_with_checks
 from sklearn.utils.validation import NotFittedError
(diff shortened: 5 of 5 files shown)


grovduck:
Hey @aazuspan, beautiful PR.  I like this design quite a bit and it feels really natural.  To your questions:

> What do you think about having `NamedFeatureArray` in `_base`?

I think `_base` seems perfectly appropriate unless there is some need for a `_utils` module.  Knowing myself, once I have a utilities type module, I throw everything that doesn't fit elsewhere in there(!), so probably not a good habit to get into.  It looks like `sklearn` has a `utils` package that has individual modules in there, so I guess it could be `sknnr/utils/_named_feature_array.py` or something like that, but I think that might be slicing it too thinly.  I'm totally fine with keeping it in `_base` if you are.

> This moves `predict` out of the individual estimators and into `TransformedKNeighborsMixin`.

Yes, let's do this (I love how clean those top-level estimators are now!), although I think there are a couple of scenarios to consider.  The first is an estimator like RF-NN which will use a "distance" measure of node similarity, where the distance between a pixel and plot is given by `(1 - (number of shared terminal nodes across tree))`.  So this ceases to be a KNeighborsRegressor anymore and won't fit this pattern anyway.

The second is a bit more nuanced.  For `Euclidean` and `Mahalanobis`, the transformations are done on the input covariates themselves and scaled in this way, i.e. the "axes" in the multivariate neighbor space is still associated with the covariates.  In this way, it doesn't make sense to eliminate any of the axes when neighbor finding.  However, both `MSN` and `GNN` (and a few others that we might consider implementing) create axes as *linear combinations* of the input covariates, so they can also serve as dimension reduction tools.  For example, we typically will run `GNN` using only a subset of the axes - this could either be a set number or a proportion of the cumulative sum of the eigenvalues.  I think this would be one or more hyperparameters of these methods and would default to the full set of axes.  And I *think* the dimension reduction can still happen in `fit` on these methods, e.g. 
```
# Untested
class GNNRegressor(IDNeighborsRegressor, TransformedKNeighborsMixin):
    def __init__(self, num_cca_axes=None):
        self.num_cca_axes = num_cca_axes
    def fit(self, X, y, spp=None):
        # CCATransformer will be responsible for returning the correct number of axes
        self.transform_ = CCATransformer(self.num_cca_axes).fit(X, y=y, spp=spp)
        return super().fit(X, y)
```
leaving `TransformedKNeighborsMixin.predict` intact.  I haven't experimented with this yet, so I may not be considering everything yet.

> To keep things simple, I just wrote a manual test for `feature_names_in_`

Totally understandable.  Like you, I imagine #21 will take a while to get right and nice to have tests in place for this one for now.  Especially because it didn't look like that was part of the "natural" suite of tests.

> One more question, what do you think about having a `_transform` method on objects with a `transform_` attribute. Is this too confusing? Maybe the method should be renamed to something like `_apply_transform`?

I do like the `_apply_transform` alternative, *unless* you would rather be more explicit in the naming of the estimated attribute `transform_`.  I'm OK with either approach ... I think it's more natural to think of methods as verbs so I have a slight preference for renaming to `_apply_transform`.

aazuspan:
> It looks like sklearn has a utils package that has individual modules in there, so I guess it could be sknnr/utils/_named_feature_array.py or something like that, but I think that might be slicing it too thinly.

Yeah, I agree. That looks like a good organization to use if we end up needing a lot more utility code, but for now it's probably overkill.

> The first is an estimator like RF-NN which will use a "distance" measure of node similarity, where the distance between a pixel and plot is given by (1 - (number of shared terminal nodes across tree)). So this ceases to be a KNeighborsRegressor anymore and won't fit this pattern anyway.

This is a very interesting point... So `RFNN` will not identify neighbors in the same way as the the other `KNeighborsRegressor` estimators, but it *will* potentially generate predictions in the same way, i.e. using weighted means of nearest neighbors, right? And I assume we will need to be able to access `kneighbors` from the estimator? Is my thinking right that we effectively want it to inherit `fit` from `RandomForestRegressor` to build and train the trees, `predict` from `KNeighborsRegressor` to calculate weighted means of nearest neighbors, and `kneighbors` from a custom implementation that uses node similarity?

In any case, it sounds like this *probably* won't interact with `TransformedKNeighborsMixin`, so luckily not something we need to fully figure out yet.

> I think this would be one or more hyperparameters of these methods and would default to the full set of axes. And I think the dimension reduction can still happen in fit on these methods

Thanks for the explanation here! I think I follow the complication, and your logic of using model hyperparameters and `fit` to run the dimensionality reduction make sense to me.

> I think it's more natural to think of methods as verbs so I have a slight preference for renaming to _apply_transform.

100%, I'll make that change!

grovduck:
Looks great!  Go forward, I say!

grovduck:
> This is a very interesting point... So `RFNN` will not identify neighbors in the same way as the the other `KNeighborsRegressor` estimators, but it *will* potentially generate predictions in the same way, i.e. using weighted means of nearest neighbors, right? And I assume we will need to be able to access `kneighbors` from the estimator?

Exactly right.

>  Is my thinking right that we effectively want it to inherit `fit` from `RandomForestRegressor` to build and train the trees, `predict` from `KNeighborsRegressor` to calculate weighted means of nearest neighbors, and `kneighbors` from a custom implementation that uses node similarity?

This is such a great question and I think the answer is *complicated*.  I'll try to give a synopsis of how RF-NN works and where I see the `sklearn` estimators fitting.

1. **fit** - One or more `y` attributes (and typically multiply `X`) are used to fit different forests, one forest per each `y` attribute.  The wa they've implemented it in yaImpute, each forest is actually a *classification* problem - the `y` attributes passed are either categorical attributes (think vegetation class) or continuous attributes that are binned into classes using some classification mode (equal interval, quantile, natural breaks, etc.).  However, the actual prediction from random forests doesn't matter (this is the mind twist) because we only care about the node IDs where the references and targets land at the deepest level, so maybe it doesn't totally matter if we inherit from `RandomForestRegressor` or `RandomForestClassifier`.  To me, the `RFNN` estimator is *composed* of one or more of these estimators (has-a) rather than being one of these (is-a).  We'll be leaning on the [apply](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestClassifier.html#sklearn.ensemble.RandomForestClassifier.apply) method of each forest to return the leaf node IDs.  I imagine that we will have an estimator attribute that holds the 2-D array of node IDs of (*m* forests by *n* trees) columns by *p* reference rows as a result of fitting.
2. **kneighbors and predict** - As opposed to all other estimators that we've introduced so far which have a neighbor space greater than 1D, the distances here are strictly based on the inverse of the number of nodes that the reference and target share in common (although a twist could be that each `y` attribute - or forest - represents its own "axis" and we'd have as many axes as `y` attributes - haven't tried this.)  So I think you're absolutely right that `kneighbors` would have to come from the `RFNN` estimator because the finding isn't based on Euclidean coordinates.  But I'm trying to figure out if we can lean on `KNeighborsRegressor` if we use the *callable* weights parameter, because given weights and `y` attributes, the actual calculation of predicted attributes will be exactly the same.  At the same time, this would be a pretty trivial function (outside of `KNeighborsRegressor`) to do the predict - I think it would just be `np.average` with a weights parameter.

Obviously, I think there is still a bit of work to get this implemented and I'm not clear on the path forward yet.  But if you see clear paths, please continue to ask questions.  We can definitely lift this from here into a separate issue as well, although it may just be thoughts at this point.

aazuspan:
Very interesting, thanks for the detailed explanations! I can see why you're saving RF-NN for last given the additional complexity over the other estimators, but it sounds like you've already thought through a lot of the nuances.

>  We can definitely lift this from here into a separate issue as well, although it may just be thoughts at this point.

This is probably a good idea! I think it's possible to go overboard setting up every foreseeable future issue, but this seems like it's clearly on the roadmap and it would be good to consolidate discussion in one place.

PR #25, opened 2023-05-25
Remove TransformedKNeighborsMixin.predict to avoid double transformation

### LINKED ISSUE #23: Unwanted double transformation using TransformedKNeighborsMixin.predict
In #22, we brought `fit`, `predict`, and `kneighbors` methods under the `TransformedKNeighborsMixin` class and took them out of the estimators.  On `predict`, I believe we are unintentionally applying each estimator's transform twice.  The current code for `TransformedKNeighborsMixin.predict` is this:
```
def predict(self, X):
    """Predict using transformed feature data."""
    X_transformed = self._apply_transform(X)
    return super().predict(X_transformed)
```
This will apply the transform to the X array and then call `KNeighborsRegressor.predict`.  However, this method then in turn calls `self.kneighbors` which returns the call back to `TransformedKNeighborsMixin.kneighbors` which again applies the transform, this time to the already transformed data.  

This probably went unnoticed because we aren't explicitly testing `predict` in [`test_port.py`](https://github.com/lemma-osu/scikit-learn-knn-regression/blob/fb_add_estimators/tests/test_port.py) at present, only `kneighbors`.  Presumably the check distances would not have matched after the double transformation.

This did surface in the implementation of the `MSNRegressor` (not yet pushed) where the canonical correlation analysis transform reduces the number of features based on significant CCA axes.  This failed both the `test_estimators_support_continuous_multioutput` and `test_estimators_support_dataframes` test because on the first transform pass, we subset the `X` array down to fewer features, but then try to pass this reduced dimension array back to `StandardScalerWithDOF` which was originally fit with the full number of features.

If this diagnosis seems correct, I believe the solution is to just remove the `TransformedKNeighborsMixin.predict` method.  Any call to an estimator's `predict` gets routed to `KNeighborsRegressor.predict` which in turn will call `TransformedKNeighborsMixin.kneighbors` and will do the appropriate transformation before neighbor finding.  I have tested locally with this approach (and on the new `MSNRegressor` estimator) and it passes all current tests.

grovduck:
Resolves #23.  The discussion there gives good background and an explanation of this fix.

### DIFF
--- a/src/sknnr/_base.py
+++ b/src/sknnr/_base.py
@@ -54,11 +54,6 @@ class TransformedKNeighborsMixin(KNeighborsRegressor):
         X_transformed = self._apply_transform(X)
         return super().fit(X_transformed, y)
 
-    def predict(self, X):
-        """Predict using transformed feature data."""
-        X_transformed = self._apply_transform(X)
-        return super().predict(X_transformed)
-
     def kneighbors(self, X=None, n_neighbors=None, return_distance=True):
         """Return neighbor indices and distances using transformed feature data."""
         X_transformed = self._apply_transform(X) if X is not None else X


aazuspan:
LGTM!

PR #31, opened 2023-05-31
Simplify fit parameters for `CCATransformer` and `CCorATransformer`

### LINKED ISSUE #30: Simplify fit parameters for `CCATransformer` and `CCorATransformer`?
While working on #20 I noticed that `CCATransformer` and `CCorATransformer` both specify `y` as an optional parameter but won't actually work if it is `None` (unless the optional `spp` is provided instead). To reproduce:

```python
from sklearn.datasets import load_linnerud
from sknnr.transformers import CCATransformer, CCorATransformer

X, y = load_linnerud(return_X_y=True)
# AxisError: axis 1 is out of bounds for array of dimension 0
CCATransformer().fit(X)
```

My first thought was that we should throw a `ValueError` if both `y` and `spp` are `None`, but now that I think about it I'm wondering if those transformers should instead just take a required `y` parameter and let the estimator handle the selection of fitting data. So for example, `GNNRegressor.fit` would pass either `y` or `spp`, but not both, to `CCATransformer.fit`. 

Currently, `CCATransformer` really only takes an `spp` parameter because the estimator that uses it might pass that, which is probably a bit backwards and could lead to some confusion, as well as some unnecessary complexity. Making the estimators responsible for choosing the right fitting data would remove a couple parameters and the need for special error handling since both estimators already require `y` parameters.

Let me know what you think, @grovduck.

aazuspan:
This will close #30 by:

1. Making `y` a required parameter for fitting `CCATransformer` and `CCorATransformer`.
2. Making `GNNRegressor` and `MSNRegressor` choose the appropriate fitting data instead of the transformer.
3. Renaming `spp` to `y_fit` throughout.

### DIFF
--- a/src/sknnr/_gnn.py
+++ b/src/sknnr/_gnn.py
@@ -3,6 +3,7 @@ from .transformers import CCATransformer
 
 
 class GNNRegressor(IDNeighborsRegressor, TransformedKNeighborsMixin):
-    def fit(self, X, y, spp=None):
-        self.transform_ = CCATransformer().fit(X, y=y, spp=spp)
+    def fit(self, X, y, y_fit=None):
+        y_fit = y_fit if y_fit is not None else y
+        self.transform_ = CCATransformer().fit(X, y=y_fit)
         return super().fit(X, y)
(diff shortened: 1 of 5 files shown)


aazuspan:
> Let's get this one merged first as I think it will have some breaking changes in https://github.com/lemma-osu/scikit-learn-knn-regression/pull/32 with the spp vs y_fit parameter.

Sounds good, thanks!

PR #32, opened 2023-05-31
Test all estimators' predict method for Moscow test data

### LINKED ISSUE #26: Test accuracy of `predict` methods
Currently, we test the accuracy of `fit` and `kneighbors` against `yaImpute` in [test_port.py](https://github.com/lemma-osu/scikit-learn-knn-regression/blob/fb_add_estimators/tests/test_port.py), but we don't yet test the accuracy of `predict`, which makes it easy for small mistakes that affect results but not functionality (like #23) to go unnoticed.

As [discussed in that issue](https://github.com/lemma-osu/scikit-learn-knn-regression/issues/23#issuecomment-1563472861), we should add some tests that compare prediction results of our estimators to those generated with `yaImpute`.

grovduck:
We exported data from yaImpute for the five core estimators at k=5 for both unweighted (equal-weighting to all neighbors) and weighted (neighbors are weighted by their inverse distance) predictions of the y attributes.  The tests assert that the predict method in each estimator is returning the same values.  Note that the constructor of each estimator uses a custom function (yaimpute_weights) for the weights parameter in order to match the weighting scheme of yaImpute.

This commit fixes #26.


aazuspan:
Looks great, I'm excited that we're finally able to confirm accuracy of predictions!

Once the test params that I broke are fixed (sorry!), looks good to merge.

Issue #33, opened 2023-05-31
Support use of reduced number of axes in CCorATransformer, CCATransformer


grovduck:
We've touched on this in a few previous issues (#20, #22) but for estimators that perform dimensionality reduction and create "axes" that are linear combinations of their input covariates (i.e. `CCorATransformer`, `CCATransformer`), we want the flexibility that a user can define the number of axes to use from these estimators.  Typically, these estimators will create the number of axes *up to* the number of `X` variables provided.  For example, `CCorATransformer` uses a test internally to reduce the number of axes down based on the statistical significance of the axes.

We need to solve the following problems with this issue:
* Allow the user to specify the number of axes to keep.  This would likely be done through the estimator's `__init__` and would default to `None`.
* After the model is fit, use the *minimum* value from either the user-specified value or from the number of axes calculated during model fit.
* If the number of axes is reduced, possibly modify and/or normalize the axis weightings to account for this.  Each estimator handles this differently.
* Use the updated number of axes and axis weightings in `transform`.
* For these estimators' "wrappers" (i.e. `MSNRegression` and `GNNRegression`), test that `predict` and `kneighbors` return the expected values.  Because yaImpute doesn't support a user-defined option for fewer axes, we will need to create new test files that implement the expected behavior and produce these results.

grovduck:
Resolved via #40

PR #40, opened 2023-06-28
Allow for user to specify reduced number of components (axes) #33

grovduck:
This feature introduces a new mixin (`ComponentReducerMixin`) to handle dimension reduction for transformers which create "axes" as linear combinations of their `X` features (presently `CCATransformer` and `CCorATransformer`).  For these transformers, there is an upper limit on the number of axes that can be used, either by matrix rank or using tests to determine significance of axes.  The user can specify a hyperparameter (`n_components`) on these transformers to use fewer than the maximum number of axes, which in turn changes the n-dimensional space for NN finding.

A few relevant changes:
- `CCA` and `CCorA` ordination classes now have a `max_components` property to ensure consistency so that either can be used in conjunction with `ComponentReducerMixin`
- `CCA` has been refactored to combine the `coefficients` and `axis_weights` into a single matrix, which has been moved into the `projector` method (similar to `CCorA`).
- Both `CCA` and `CCorA`'s projector method now accommodates a parameter for `n_components`
- Tests have been added to ensure that transformed `X` data has the correct number of components and raise errors if the user specified value is outside the range [0 - `max_components`].


grovduck:
@aazuspan, I think this is still pretty rough, but wanted to get some feedback from you in terms of design and specifically the tests.  I'll add in some specific comments inlined with the code, but here are the general points:

* There is a new mixin (`ComponentReducerMixin`) that handles the dimension reduction.  I followed the naming of sklearn's [`PCA`](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html) and use `n_components` to describe the number of axes to retain.  Note that I'm much more used to using the term "axes" instead of "components", but did so to remain consistent.  This mixin doesn't do much, but I've introduced it to reduce code duplication for both `CCATransformer` and `CCorATransformer`.  

* I've tweaked the `CCA` and `CCorA` classes to make them more consistent, specifically adding a `max_components` property and a `projector` method that can take `n_components`.  This takes some of the logic out of their respective transformers.

* The tests are pretty janky.  I've got two tests in there to make sure that a transformed `X` is reduced in dimensionality correctly and that it throws an error when the user-specified value is out of range.  But I don't like the parameterization that I have for `n_components` in either test.  For the first, I've used `[None, 0, 5]` because all of those are valid values for this specific test data.  However, because of axis significance tests in `CCorA`, it already reduces the number of axes from 28 to 5.  Do you think we should first retrieve this value (e.g. `CCorA(X, y).max_components`) in the test before running the test.  How could you parameterize such an expression?  For the second, I'm just testing a negative value and an absurdly high value which both fall out of range, but not psyched about the constant values again.

* I haven't yet created the data to test the *estimators* with reduced transformers, but I'd like to include those as part of this PR.  These would be porting tests again, but I'll need to do this outside of yaImpute.

src/sknnr/transformers/_base.py
```diff
@@ -11,3 +11,25 @@ def fit(self, X, y=None, sample_weight=None):
...
+
+class ComponentReducerMixin:
+    """
+    Mixin for transformers that allow reduction of the number of components.
+    """
+
+    def __init__(self, n_components=None):
+        self.n_components = n_components
+
+    def set_components(self, ordination_obj):
```
grovduck, line 24: I haven't yet done any typing on these functions.  `ordination_obj` here is an instance of `CCA` or `CCorA`, so probably just needs a union type?

src/sknnr/transformers/_base.py
```diff
@@ -11,3 +11,25 @@ def fit(self, X, y=None, sample_weight=None):
...
+            self.n_components
+            if self.n_components is not None
+            else ordination_obj.max_components
+        )
+        if not 0 <= n_components <= ordination_obj.max_components:
+            raise ValueError(
+                f"n_components={n_components} must be between 0 and "
+                f"{ordination_obj.max_components}"
+            )
+        self.n_components_ = n_components
```
grovduck, line 35: Is it "suprising" to be setting an estimator attribute within the mixin.  I feel like it's alright, but we could also just return it and set it in `CCATransformer` and `CCorATransformer`.

src/sknnr/transformers/_cca_transformer.py
```diff
@@ -3,13 +3,14 @@
...
+from . import ComponentReducerMixin
 from ._cca import CCA
 
 
-class CCATransformer(TransformerMixin, BaseEstimator):
+class CCATransformer(ComponentReducerMixin, TransformerMixin, BaseEstimator):
     @property
     def _n_features_out(self):
-        return self.cca_.eigenvalues.shape[0]
+        return self.n_components_
```
grovduck, line 13: This is obviously silly to just return another attribute with this property, but I think things would break with `_n_features_out` not defined, correct?

src/sknnr/transformers/_cca_transformer.py
```diff
@@ -23,15 +24,14 @@ def fit(self, X, y):
 
         X, y = np.asarray(X), np.asarray(y)
         self.cca_ = CCA(X, y)
+        self.set_components(self.cca_)
+        self.projector_ = self.cca_.projector(n_components=self.n_components_)
```
grovduck, line 28: We don't necessarily need to set an estimator attribute here, we could instead just do this in `transform` instead.  Preference?

tests/test_transformers.py
```diff
@@ -89,3 +89,23 @@ def test_transformers_raise_notfitted_transform(transformer):
     X, y = load_moscow_stjoes(return_X_y=True)
     with pytest.raises(NotFittedError):
         transformer().transform(X)
+
+
+@pytest.mark.parametrize("transformer", [CCATransformer, CCorATransformer])
+@pytest.mark.parametrize("n_components", [None, 0, 5])
```
grovduck, line 95: Janky!!!

tests/test_transformers.py
```diff
@@ -89,3 +89,23 @@ def test_transformers_raise_notfitted_transform(transformer):
     X, y = load_moscow_stjoes(return_X_y=True)
     with pytest.raises(NotFittedError):
         transformer().transform(X)
+
+
+@pytest.mark.parametrize("transformer", [CCATransformer, CCorATransformer])
```
grovduck, line 94: Do you want to set a module level list like `TEST_TRANSFORMERS` or do you prefer this syntax?

src/sknnr/transformers/_base.py
```diff
@@ -11,3 +11,25 @@ def fit(self, X, y=None, sample_weight=None):
         scaler = super().fit(X, y, sample_weight)
         scaler.scale_ = np.std(np.asarray(X), axis=0, ddof=self.ddof)
         return scaler
+
+
+class ComponentReducerMixin:
```
aazuspan, line 16: I like this mixin solution! Very clean.

src/sknnr/transformers/_cca.py
```diff
@@ -187,6 +191,14 @@ def coefficients(self):
         right = Q.T @ self._U
         return np.linalg.lstsq(R, right, rcond=None)[0]
 
+    def projector(self, n_components=None):
+        if n_components is None:
+            return self.coefficients @ self.axis_weights
```
aazuspan, line 196: What do you think about replacing this early return with `n_components = self.max_components`? In other words:

```python
def projector(self, n_components=None):
        if n_components is None:
            n_components = self.max_components

        return (
            self.coefficients[:, :n_components]
            @ self.axis_weights[:n_components, :n_components]
        )
```

I find that a little more readable because the same return will be used in both cases, but maybe the fact that the indexing would have no effect outweighs that benefit?

src/sknnr/transformers/_ccora.py
```diff
@@ -129,5 +129,12 @@ def n_vec(self):
...
 
     @property
-    def projector(self):
-        return self.x_coef[:, : self.n_vec] @ np.diag(self.cancorr[: self.n_vec])
+    def max_components(self):
+        return self.n_vec
+
+    def projector(self, n_components=None):
+        if n_components is None:
+            return self.x_coef[:, : self.max_components] @ np.diag(
```
aazuspan, line 137: Similar to my comment on `ConstrainedOrdination.projector`, but in this case I think there's no downside to replacing the early return with `n_components = self.max_components` and using the same return for both cases, right? i.e.

```python
    def projector(self, n_components=None):
        if n_components is None:
            n_components = self.max_components
        return self.x_coef[:, :n_components] @ np.diag(self.cancorr[:n_components])
```

tests/test_transformers.py
```diff
@@ -89,3 +89,23 @@ def test_transformers_raise_notfitted_transform(transformer):
...
+def test_transformers_n_components(transformer, n_components):
+    """Test that n_components is handled correctly."""
+    X, y = load_moscow_stjoes(return_X_y=True)
+    t = transformer(n_components=n_components).fit(X, y)
+    assert t._n_features_out == t.n_components_
+    assert t.transform(X).shape[1] == t.n_components_
+
+
+@pytest.mark.parametrize("transformer", [CCATransformer, CCorATransformer])
+@pytest.mark.parametrize("n_components", [-1, 1000])
```
aazuspan, line 105: > For the second, I'm just testing a negative value and an absurdly high value which both fall out of range, but not psyched about the constant values again.

I get the uneasiness, but I think the choice of constants is a little clearer here than in the first test, especially since the docstring basically explains that these are out of range. I'm okay with this as-is.

aazuspan (commented):
This is looking great, @grovduck! I left some relatively minor comments, but I think overall this is looking like a really clean design. I think the mixin approach worked great.

> I followed the naming of sklearn's [PCA](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html) and use n_components to describe the number of axes to retain. Note that I'm much more used to using the term "axes" instead of "components", but did so to remain consistent.

I think this was the right call!

> I haven't yet created the data to test the estimators with reduced transformers, but I'd like to include those as part of this PR. These would be porting tests again, but I'll need to do this outside of yaImpute.

Great, it makes sense to me to include that in this PR too. I assume that means the ability to specify `n_components` when instantiating estimators would happen here, too?

I'll spend a little more time thinking about test cases tomorrow and let you know if anything else occurs to me.

src/sknnr/transformers/_base.py
grovduck: Absolutely, this is the right call.  I went with `ordination_` just to be a bit more explicit.

src/sknnr/transformers/_cca_transformer.py
grovduck: Sweet, that works.  I had in the back of my brain that `_n_features_out` was critical, but it looks to only show up with the [`ClassNamePrefixFeaturesOutputMixin`](https://scikit-learn.org/stable/modules/generated/sklearn.base.ClassNamePrefixFeaturesOutMixin.html) that we had played around with, but ultimately wrote our own.  Thanks for the good thought!

src/sknnr/transformers/_cca_transformer.py
grovduck: Agreed and fixed.  Now using `self.ordination_.projector(n_components=self.n_components_)` in `transform`.

tests/test_transformers.py
grovduck: I'll sit on this for the evening as well, although I think your last thought is a good one.

tests/test_transformers.py
grovduck: Sounds good to me.  I've gone with `TEST_ORDINATION_TRANSFORMERS` for now.  "ordination" was the best term that I could think of to differentiate these two transformers from the ones that retain all of their covariates.

tests/test_transformers.py
```diff
@@ -89,3 +89,23 @@ def test_transformers_raise_notfitted_transform(transformer):
...
+def test_transformers_raise_out_of_range_n_components(transformer, n_components):
+    """Attempting to call fit with an out of range value of n_components
+    should raise."""
+    X, y = load_moscow_stjoes(return_X_y=True)
+    with pytest.raises(ValueError):
```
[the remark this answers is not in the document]
grovduck: Ooh, cool!  I've not done one of those before (although I obviously should have looked at your examples).  I used this match:
```
n_components=-?\d+ must be between 0 and \d+
```
Sound OK?

src/sknnr/transformers/_base.py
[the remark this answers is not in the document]
grovduck: Good catch.  Fixed.

tests/test_transformers.py
```diff
@@ -89,3 +89,23 @@ def test_transformers_raise_notfitted_transform(transformer):
...
         transformer().transform(X)
+
+
+@pytest.mark.parametrize("transformer", [CCATransformer, CCorATransformer])
+@pytest.mark.parametrize("n_components", [None, 0, 5])
+def test_transformers_n_components(transformer, n_components):
+    """Test that n_components is handled correctly."""
+    X, y = load_moscow_stjoes(return_X_y=True)
+    t = transformer(n_components=n_components).fit(X, y)
+    assert t._n_features_out == t.n_components_
```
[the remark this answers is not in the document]
grovduck: Thank you!  I was struggling with how to do this cleanly *and* include `None` but could not figure out a good way to do it.  Ignoring `None` on this specific test is the right call.

src/sknnr/transformers/_cca.py
grovduck: I agree, much cleaner!

src/sknnr/transformers/_ccora.py
grovduck: Ditto here!

grovduck:
As always, a great review catching lots of little errors.  Still need to work on estimator tests, but I think I've addressed most of your comments.

tests/test_transformers.py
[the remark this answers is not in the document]
aazuspan: Looks great! I would've probably been lazy and just matched against "must be between 0 and", but the more specific regex is definitely better!

src/sknnr/transformers/_cca.py
```diff
@@ -193,7 +193,7 @@ def coefficients(self):
 
     def projector(self, n_components=None):
         if n_components is None:
-            return self.coefficients @ self.axis_weights
+            n_components = self.max_components
```
aazuspan, line 196: I didn't notice until I ran test coverage and saw this is uncovered, but this condition won't ever happen during normal usage because `n_components` will always be passed *after* it goes through `ComponentReducerMixin.set_n_components`, which effectively runs the same check for `None` values. 

I'm fine leaving this in if you think there are possible cases where a user might want to access this directly, but if not then maybe we don't need the check at all and `n_components` can be a required param?

aazuspan (commented):
The changes look great! I just noted a few things that I missed on my first pass.

grovduck:
> Great, it makes sense to me to include that in this PR too. I assume that means the ability to specify `n_components` when instantiating estimators would happen here, too?

@aazuspan, I've got the bones of this for `GNNRegressor` ready to go - both test data and the required additions to the estimator.  Before I dump a bunch of new (reduced) test files in, I thought we should discuss what makes sense from a testing perspective.  I've run GNN with n=8 components (using `pynnmap`) and have those test files ready:
```
gnn_moscow_ref_distances_k5_c8.csv
gnn_moscow_ref_neighbors_k5_c8.csv
```
Do we want to change code to accommodate these new files.  Is this a one-off for testing or will we want to keep these data files around?  I like the "reassurance" that we're getting the correct values, but I don't know how long we want to retain all the porting info?  Do you have any opinions on this?

src/sknnr/transformers/_cca_transformer.py
```diff
@@ -8,30 +8,25 @@
 
 
 class CCATransformer(ComponentReducerMixin, TransformerMixin, BaseEstimator):
-    @property
-    def _n_features_out(self):
-        return self.n_components_
-
     def get_feature_names_out(self) -> NDArray:
```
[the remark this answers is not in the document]
grovduck: Brilliant!  I like this idea, but could use your help.  I've taken `get_feature_names_out` of both `CCATransformer` and `CCorATransformer` and put it into `ComponentReducerMixin` like this:
```
    def get_feature_names_out(self) -> NDArray:
        check_is_fitted(self, ["ordination_", "n_components_"])
        return np.asarray(
            [
                f"{self.ordination_.__class__.__name__.lower()}{i}"
                for i in range(self.n_components_)
            ],
            dtype=object,
        )
```
One issue, two questions:
1. mypy isn't passing pre-commit based on this message: `error: "ComponentReducerMixin" has no attribute "ordination_"  [attr-defined]`.  Because `ordination_` is getting set in the `fit` method of the transformers, `ComponentReducerMixin` doesn't see it.  I feel like you've solved this issue before - do I have to declare it somewhere in `ComponentReducerMixin`?
2. Do you want a custom error message on `check_is_fitted` or will the default one work: `“This %(name)s instance is not fitted yet. Call ‘fit’ with appropriate arguments before using this estimator.”`?
3. How do you show co-authorship on a commit?  I've seen that you've done this, but I wasn't sure how to do this.  I'm getting quite a bit of help here from you on this PR!

src/sknnr/transformers/_cca.py
grovduck: Great question.  When I first wrote these classes, it was basically code dumps for the `CCA` and `CCorA` classes from stuff I had already done and then `CCATransformer` and `CCorATransformer` were meant to be thin layers on top to conform to scikit-learn.  While one could certainly call `CCA` or `CCorA` directly, I don't think that would be a typical use case.  For now, I like your suggestion and will make this a required param.  If it comes back to bite us, it's an easy transition back to this form.

src/sknnr/transformers/_cca_transformer.py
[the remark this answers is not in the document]
aazuspan: >  I feel like you've solved this issue before - do I have to declare it somewhere in ComponentReducerMixin?

Yep, here's that line in `TransformedKNeighborsMixin`: https://github.com/lemma-osu/scikit-learn-knn-regression/blob/55e079d913f8deb939b57bdb8cad3bedcae14c17/src/sknnr/_base.py#L44

Since the ordinations don't share a superclass, I guess the declaration should be `ordination_: CCorA | CCA`? Or that could be declared as a module-level type, e.g. `Ordination = CCorA | CCA`. I could go either way.

This is probably too big of an issue to tackle here, but I wonder if those could/should be refactored to share an `Ordination` superclass? I'm not sure if they really share enough functionality for that to make sense.

> Do you want a custom error message on check_is_fitted or will the default one work: “This %(name)s instance is not fitted yet. Call ‘fit’ with appropriate arguments before using this estimator.”?

The default's fine with me - that's what `PCA.get_feature_names_out` does.

> How do you show co-authorship on a commit? I've seen that you've done this, but I wasn't sure how to do this. I'm getting quite a bit of help here from you on this PR!

Don't worry about it! I think that only happens when someone suggests a specific code change that gets committed :)

aazuspan:
> Do we want to change code to accommodate these new files. Is this a one-off for testing or will we want to keep these data files around? I like the "reassurance" that we're getting the correct values, but I don't know how long we want to retain all the porting info? Do you have any opinions on this?

This is a great question. I originally thought the port tests would be temporary, but I agree with you that it's nice to know that everything is staying accurate, as well as functional, so I'm fully in favor of keeping them around in some form or another. 

Eventually, I suppose we will reach a point where we're confident everything is working and could theoretically test against previous versions of `sknnr`, although I'm not sure exactly how that would happen in practice. 

For now, I guess I would lean towards integrating `n_components` into our testing system. Maybe it could just be another argument to `tests.datasets.load_moscow_stjoes_results` that gets appended (if it's not `None`) to the file names?

src/sknnr/transformers/_cca.py
aazuspan: Cool, makes sense to me!

src/sknnr/transformers/_cca_transformer.py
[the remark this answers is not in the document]
grovduck: > This is probably too big of an issue to tackle here, but I wonder if those could/should be refactored to share an `Ordination` superclass? I'm not sure if they really share enough functionality for that to make sense.

Oh, absolutely!  And there are a couple more ordination techniques that are even more closely related to CCA that would make this a no-brainer.  Famous last words, but I fully intend to refactor those two classes and find some commonalities between them and having an `Ordination` base class for these makes a lot of sense.  Not for this PR, but I will start an issue.

> Since the ordinations don't share a superclass, I guess the declaration should be `ordination_: CCorA | CCA`? Or that could be declared as a module-level type, e.g. `Ordination = CCorA | CCA`. I could go either way.

I went with the latter with the idea that if we *do* create a base class, we'll just remove this line and import it from elsewhere.  Two other comments:

1. You can probably see that I had a failing test, because I tried to use `CCorA | CCA` syntax rather than `Union[CCorA, CCA]` syntax.  I tried with `from __future__ import annotations`, but that doesn't cover the union operator, correct?
2. Check me on this, but because we now have `ordination_` as a class attribute, it makes no sense to pass it to `check_is_fitted`, because it is only checking for the presence of that estimator attribute, correct?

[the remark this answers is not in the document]
aazuspan: > I went with the latter with the idea that if we do create a base class, we'll just remove this line and import it from elsewhere.

Good thinking!

>  I tried with from __future__ import annotations, but that doesn't cover the union operator, correct?

Interesting, the tests failed with that import in place? I just switch from `Union` to `|` for my type annotations in [sankee](https://github.com/aazuspan/sankee/blob/1242246d25b8e15918bea8faec735270c316dd93/sankee/sampling.py#L16) and the `3.8` tests pass there as long as I have the `__future__` import. It would be nice to have `tox` set up to be able to reproduce these version-specific bugs locally...

> Check me on this, but because we now have ordination_ as a class attribute, it makes no sense to pass it to check_is_fitted, because it is only checking for the presence of that estimator attribute, correct?

Yeah, good point. Unless there's some hypothetical case where it could be have some fitted attributes but not others (not a possibility here), I don't think there's any particular reason to specify those parameters.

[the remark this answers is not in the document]
grovduck: > Interesting, the tests failed with that import in place? I just switch from `Union` to `|` for my type annotations in [sankee](https://github.com/aazuspan/sankee/blob/1242246d25b8e15918bea8faec735270c316dd93/sankee/sampling.py#L16) and the `3.8` tests pass there as long as I have the `__future__` import. It would be nice to have `tox` set up to be able to reproduce these version-specific bugs locally...

I hadn't explicitly tried it before, but I just committed and pushed with the union operator in there and it [does bonk](https://github.com/lemma-osu/scikit-learn-knn-regression/actions/runs/5417534136/jobs/9848606292?pr=40).  Maybe I'm doing something wrong?  (I went ahead and reverted to `Union[CCA, CCorA]`)

> Unless there's some hypothetical case where it could be have some fitted attributes but not others (not a possibility here), I don't think there's any particular reason to specify those parameters

I want to make sure I understand.  Currently, I'm still checking for `n_components_`, e.g.
```
check_is_fitted(self, "n_components_")
```
but are you saying we should just not specify the attributes parameter, in which case it will just verify that a single estimator attribute is present?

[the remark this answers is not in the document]
aazuspan: > I hadn't explicitly tried it before, but I just committed and pushed with the union operator in there and it [does bonk](https://github.com/lemma-osu/scikit-learn-knn-regression/actions/runs/5417534136/jobs/9848606292?pr=40). Maybe I'm doing something wrong? (I went ahead and reverted to Union[CCA, CCorA])

No, you were right! Thanks for checking. Apparently there's a difference in behavior (at least in 3.8) between type annotations within a function (like I used in sankee) and type aliases (like we're using here), we I think we're stuck with `Union`. Here's a relevant [snippet from the mypy docs](https://mypy.readthedocs.io/en/stable/runtime_troubles.html#future-annotations-import-pep-563) that vaguely mentions this:

image

As you said before, typing is hard!

> but are you saying we should just not specify the attributes parameter, in which case it will just verify that a single estimator attribute is present?

That was what I was saying, but your solution of checking for `n_components_` is more explicit and is consistent with the [check in PCA](https://github.com/scikit-learn/scikit-learn/blob/364c77e047ca08a95862becf40a04fe9d4cd2c98/sklearn/base.py#L944), so your instinct was right!

tests/test_transformers.py
grovduck: I ended up putting a comment in here explaining the use of "5" as a test case

grovduck:
> For now, I guess I would lean towards integrating `n_components` into our testing system. Maybe it could just be another argument to `tests.datasets.load_moscow_stjoes_results` that gets appended (if it's not `None`) to the file names?

OK, I've got a straw man for testing in there using this suggestion.  At present, I'm only testing against a 3-component version of both `GNNRegressor` and `MSNRegressor`.  The test files for `GNNRegressor` were created "independently" through `pynnmap`, so I'm fairly confident that tests are working as intended.  For `MSNRegressor`, I "cheated" by actually running `sknnr` with reduced components to get the neighbor, distance, and predicted files, so there is no mystery why those tests are passing.  I felt a little dirty about doing it this way, but I felt that it was going to take a while to do through `yaImpute` and I don't have it set up in `pynnmap`.  And it will be good for future tests as well.  I'm 99.9% sure that the test files I produced really are the correct answer (famous last words ...)

A couple of other things to note:

1. `GNNRegressor` and `MSNRegressor` are looking way too similar to not be refactored into a common base class (at some point).  We could easily abstract out the transformer in the base class and specify these in the derived classes.  Probably good to tackle when I do the `CCA` and `CCorA` reworking.
2. In `test_port.py`, functions `test_kneighbors_reduced_components` and `test_predict_reduced_components`, I have `n_components = 3` hard-wired in there.  Should we parametrize this even if we don't have other `n_components` to test?

Sorry this PR is turning into such a beast - I appreciate your help!

grovduck:
> EDIT: [pytest-regressions](https://github.com/ESSS/pytest-regressions) and [syrupy](https://github.com/tophat/syrupy) both look like interesting tools that might solve this problem for us in the future by testing against automatically generated results...

Oooo, super cool.  I'd want to play around with these packages to fully understand them, but they seem totally applicable for our use.  Nice find!

aazuspan (approved):
The changes look great - no notes from me! At first glance I was hoping we could parameterize `n_components` as `[None, 3]` and run it with all estimators to avoid duplicating those tests, but I'm guessing you probably already thought of that. I poked around a little while and couldn't come up with any good solution (apparently there's no way to [silently skip a test](https://github.com/pytest-dev/pytest/issues/3730) in pytest), so I'd say go ahead and merge if you're ready!

> For MSNRegressor, I "cheated" by actually running sknnr with reduced components to get the neighbor, distance, and predicted files, so there is no mystery why those tests are passing. I felt a little dirty about doing it this way, but I felt that it was going to take a while to do through yaImpute and I don't have it set up in pynnmap. And it will be good for future tests as well. I'm 99.9% sure that the test files I produced really are the correct answer (famous last words ...)

Good to know! If you're confident in the answers, that's good enough for me :)

> GNNRegressor and MSNRegressor are looking way too similar to not be refactored into a common base class (at some point). We could easily abstract out the transformer in the base class and specify these in the derived classes. Probably good to tackle when I do the CCA and CCorA reworking.

I had the same thought, but I'm happy to postpone that for a future PR.

> In test_port.py, functions test_kneighbors_reduced_components and test_predict_reduced_components, I have n_components = 3 hard-wired in there. Should we parametrize this even if we don't have other n_components to test?

Nah, I think this is a reasonable approach. If we eventually update the tests to use dynamically generated results with one of those tools we could parameterize over a range of `n_components`, but this looks good for now!

> Oooo, super cool. I'd want to play around with these packages to fully understand them, but they seem totally applicable for our use. Nice find!

Yeah, I'm pretty excited about those! Just glancing through the docs, it sounds like that should simplify out test dataset system a lot.

grovduck:
> At first glance I was hoping we could parameterize `n_components` as `[None, 3]` and run it with all estimators to avoid duplicating those tests, but I'm guessing you probably already thought of that. I poked around a little while and couldn't come up with any good solution (apparently there's no way to [silently skip a test](https://github.com/pytest-dev/pytest/issues/3730) in pytest), so I'd say go ahead and merge if you're ready!

Again, you give me way too much credit.  I hadn't thought of removing that duplication, but it makes sense.  You're way ahead of me here, but is this what you were thinking?
```
@pytest.mark.parametrize(
    "result", ESTIMATOR_RESULTS.items(), ids=ESTIMATOR_RESULTS.keys()
)
@pytest.mark.parametrize("n_components", [None, 3], ids=["full", "reduced"])
def test_kneighbors(result, n_components):
    """Test that the ported estimators identify the correct neighbors and distances."""
    method, estimator = result

    if method not in ORDINATION_ESTIMATOR_RESULTS.keys() and n_components is not None:
        pytest.skip("These methods do not support variable n_components.")
    dataset = load_moscow_stjoes_results(method=method, n_components=n_components)

    hyperparams = (
        {"n_neighbors": 5}
        if method not in ORDINATION_ESTIMATOR_RESULTS.keys()
        else {"n_neighbors": 5, "n_components": n_components}
    )

    est = estimator(**hyperparams).fit(
        dataset.X_train, dataset.y_train
    )

    dist, nn = est.kneighbors(return_dataframe_index=True)
    assert_array_equal(nn, dataset.ref_neighbors)
    assert_array_almost_equal(dist, dataset.ref_distances, decimal=3)

    dist, nn = est.kneighbors(dataset.X_test, return_dataframe_index=True)
    assert_array_equal(nn, dataset.trg_neighbors)
    assert_array_almost_equal(dist, dataset.trg_distances, decimal=3)
```
It's a little verbose and has too much logic, but I'm happy to change to this if you like this form better.  It's not silently skipping (as you say), but is that OK?  I read through that thread you sent and it looks like the optimal solution would be to filter out the combinations that don't make sense (e.g. `("raw", 3)`) as part of collection rather than in the test itself.

aazuspan:
Yes, that's almost exactly the solution I came up with. It feels odd to me to have tests that will always be skipped, but maybe that's not a big deal? It is nice not to duplicate the logic of those two tests.

> it looks like the optimal solution would be to filter out the combinations that don't make sense (e.g. ("raw", 3)) as part of collection rather than in the test itself.

Correct me if I'm wrong, but my understanding is that this would require combining the two parameterized args (`result` and `n_components`) into a single arg that contains all valid combinations of `result` and `n_components`, which seems to me like it would hurt readability and flexibility. Or the other alternative seemed to be monkey patching some new functionality into `pytest`, which I also don't love.

Do you have any preference? I think I lean towards either leaving it as-is with duplicated tests or conditionally skipping the invalid tests, and I could live with either.

grovduck:
> Correct me if I'm wrong, but my understanding is that this would require combining the two parameterized args (`result` and `n_components`) into a single arg that contains all valid combinations of `result` and `n_components`, which seems to me like it would hurt readability and flexibility. Or the other alternative seemed to be monkey patching some new functionality into `pytest`, which I also don't love

In the thread you sent, I think [this post](https://github.com/pytest-dev/pytest/issues/3730#issuecomment-567142496) was the closest I saw to a suggested solution.  I'm playing around with it a bit - it looks like you use the `pytest_collection_modifyitems` hook to filter out the combinations you don't want.  Would you consider that too magic?

If that's too much, I think I'll probably keep the duplicated tests in there, just because it sounds like the consensus is that `skip` is meant to denote a test that may pass at *some point *.

grovduck:
Here's a worked example, copying most of what was in that post:
```
conftest.py
------------
import pytest


def pytest_configure(config):
    config.addinivalue_line(
        "markers",
        "uncollect_if(*, func): function to unselect tests from parametrization",
    )


def pytest_collection_modifyitems(config: pytest.Config, items: list[pytest.Item]):
    removed = []
    kept = []
    for item in items:
        m = item.get_closest_marker("uncollect_if")
        if m:
            func = m.kwargs["func"]
            if func(**item.callspec.params):
                removed.append(item)
                continue
        kept.append(item)
    if removed:
        config.hook.pytest_deselected(items=removed)
        items[:] = kept

test_port.py
------------
def uncollect_if(result, n_components):
    method, _ = result
    if n_components is not None and method not in ORDINATION_ESTIMATOR_RESULTS.keys():
        return True


@pytest.mark.uncollect_if(func=uncollect_if)
@pytest.mark.parametrize(
    "result", ESTIMATOR_RESULTS.items(), ids=ESTIMATOR_RESULTS.keys()
)
@pytest.mark.parametrize("n_components", [None, 3], ids=["full", "reduced"])
def test_kneighbors(result, n_components):
    """Test that the ported estimators identify the correct neighbors and distances."""
    method, estimator = result

    dataset = load_moscow_stjoes_results(method=method, n_components=n_components)

    hyperparams = (
        {"n_neighbors": 5}
        if method not in ORDINATION_ESTIMATOR_RESULTS.keys()
        else {"n_neighbors": 5, "n_components": n_components}
    )

    est = estimator(**hyperparams).fit(dataset.X_train, dataset.y_train)

    dist, nn = est.kneighbors(return_dataframe_index=True)
    assert_array_equal(nn, dataset.ref_neighbors)
    assert_array_almost_equal(dist, dataset.ref_distances, decimal=3)

    dist, nn = est.kneighbors(dataset.X_test, return_dataframe_index=True)
    assert_array_equal(nn, dataset.trg_neighbors)
    assert_array_almost_equal(dist, dataset.trg_distances, decimal=3)
```
with the corresponding output:
```
PS D:\code\scikit-learn-knn-regression> hatch run test:all tests/test_port.py::test_kneighbors -v
============================================================================= test session starts =============================================================================
platform win32 -- Python 3.10.1, pytest-7.3.1, pluggy-1.0.0 -- C:\Users\gregorma\AppData\Local\hatch\env\virtual\scikit-learn-knn-regression\47ek3DpI\test\Scripts\python.EXE
cachedir: .pytest_cache
rootdir: D:\code\scikit-learn-knn-regression
configfile: pyproject.toml
plugins: cov-4.0.0
collected 10 items / 3 deselected / 7 selected

tests/test_port.py::test_kneighbors[full-raw] PASSED                                                                                                                     [ 14%]
tests/test_port.py::test_kneighbors[full-euclidean] PASSED                                                                                                               [ 28%]
tests/test_port.py::test_kneighbors[full-mahalanobis] PASSED                                                                                                             [ 42%]
tests/test_port.py::test_kneighbors[full-gnn] PASSED                                                                                                                     [ 57%]
tests/test_port.py::test_kneighbors[full-msn] PASSED                                                                                                                     [ 71%]
tests/test_port.py::test_kneighbors[reduced-gnn] PASSED                                                                                                                  [ 85%]
tests/test_port.py::test_kneighbors[reduced-msn] PASSED                                                                                                                  [100%]

======================================================================= 7 passed, 3 deselected in 4.03s =======================================================================

**Comment by aazuspan** [comment]
Ah, I glossed over that post because I was a bit confused by it, to be honest. I'm not very familiar with the `pytest` hook system, but it does look like it's probably the cleanest solution, overall. It *does* feel slightly magic, but it also seems like it's idiomatic for `pytest`.

 If you want to adapt that for our use case, I'm happy with that solution!

**Comment by grovduck** [comment]
Arrgh, I'm losing steam on this PR ... A couple of blockers/annoyances:

1. I tried using the new dictionary union operator (`|`), but it fails on 3.8.  Then I tried this syntax instead `d = dict("a"=1); d.update({"b", 2})`, but now `sourcery` complains on pre-commit that I should be using the union operator.  I think I'm almost ready to ditch `sourcery` as a pre-commit because it doesn't look like you can specify options via `pyproject.toml` and because I'm using python 3.10 in development, it always wants me to incorporate the new features (same thing goes for the walrus operator in one case).

2. We can filter out the impossible combinations of (method, n_components), but in the test we also need to change the hyperpameters based on the estimator type as well.  This means we're writing the same check (i.e. `hasattr(estimator(), "n_components"`) both in the new `estimator_does_not_support_n_components` function as well as the test itself.

3. In order to use `estimator_does_not_support_n_components` in both `test_kneighbors` and `test_predict`, the function signature has to accomodate `result` and `n_components` from `test_kneighbors`, but also the `weighted` parameter from `test_predict`.  For now, the function signature looks like `def estimator_does_not_support_n_components(result, n_components, **kwargs)`, but I'm not sure that's the right approach.

I could use your feedback on these issues.  The first is a blocker for actually pushing the code and it might be easier to provide feedback on the second two after I commit/push my new changes.

**Comment by aazuspan** [comment]
> I tried using the new dictionary union operator (|), but it fails on 3.8. Then I tried this syntax instead d = dict("a"=1); d.update({"b", 2}), but now sourcery complains on pre-commit that I should be using the union operator. I think I'm almost ready to ditch sourcery as a pre-commit because it doesn't look like you can specify options via pyproject.toml and, because I'm using python 3.10 in development, it always wants me to incorporate the new features (same thing goes for the walrus operator in one case).

Wow, dict unions are new to me! I feel like I always miss out on the cool new features by maintaining backwards compatible packages... 

Back to the issue at hand, I'm open to dropping `sourcery` from our pre-commit. I think I've accepted a few corrections from it that I was ambivalent about, but it's also pretty slow to run and the fact that any new contributor would have to sign up for an account to run our checks is a little off-putting. With the lack of configuration and the Python version issues, I'd say it's doing more harm than good.

As far as *how* we remove it, I'm totally open to you just disabling it in this PR and then cleaning up the rest of the pre-commit stuff in #41. That's probably not strictly the cleanest way to do it, but I think the alternative of doing #41 first would end up blocking this for a while and probably create a headache when you have to merge in all the changes that will stem from running the `pyupgrade` rules. Unless it feels too janky to you to combine this PR with a change to the pre-commit, that's the lazy approach I would suggest. 

As recommended, I'll take a closer look at the other questions once we get to a resolution here!

**Comment by grovduck** [comment]
> As far as _how_ we remove it, I'm totally open to you just disabling it in this PR and then cleaning up the rest of the pre-commit stuff in #41. That's probably not strictly the cleanest way to do it, but I think the alternative of doing #41 first would end up blocking this for a while and probably create a headache when you have to merge in all the changes that will stem from running the `pyupgrade` rules. Unless it feels too janky to you to combine this PR with a change to the pre-commit, that's the lazy approach I would suggest.

Excellent, all `sourcery` code is currently commented out.  With #41, I'll leave it to you for the removal (we'll probably want to remove the key as well?).  `black` did some spacing formatting on `.pre-commit-config.yaml`, so hopefully that jives with how it formats on your end.

To your earlier code suggestions:

> 1. Probably add a docstring to `pytest_collection_modifyitems` to explain what it's doing and link to the Github post.

I've done this and added a link to the specific comment in the `pytest` issue

> 2. In `conftest.py`, can we [move the marker configuration into `pyproject.toml`](https://docs.pytest.org/en/latest/how-to/mark.html#registering-marks)? Defining the config programmatically feels a little odd, outside of a plugin approach where it needs to be patched in after-the-fact.

Cool, I didn't know you could do this here, but I think I've done it correctly.

> 3. In `test_port.py`, can we rename `uncollect_if` to something specific, like `estimator_does_not_support_ncomponents` (or something shorter)? It would be even better if it could be inlined, but I imagine that would be too long to be easily readable?

I actually liked your naming suggestion for the function, so I kept it.  I didn't inline it because we can use it for both tests (I *think* you meant to inlining the function within the test itself, right?)

> 4. In the same function above, what do you think about replacing the `method not in ORDINATION_ESTIMATOR_RESULTS.keys()` check with `not hasattr(method, "n_components")`? I think that accomplishes the same thing and would allow us to get rid of `ORDINATION_ESTIMATOR_RESULTS`, but I don't have a strong feeling so let me know if there's a downside I'm not considering.

I've done this as well and now `ORDINATION_ESTIMATOR_RESULTS` is no longer needed.

Hopefully, the two questions/concerns I raised earlier will now make sense.  Thanks for keeping with me on this one - it's turned out to be a bit epic!

File: tests/test_port.py [hunk 15]
```diff
@@ -24,15 +24,28 @@ def yaimpute_weights(d):
     return 1.0 / (1.0 + d)
 
 
+def estimator_does_not_support_n_components(result, n_components, **kwargs):
+    _, estimator = result
+    if n_components is not None and not hasattr(estimator(), "n_components"):
+        return True
```
aazuspan, line 30: Refactoring nit:

```suggestion
    return n_components is not None and not hasattr(estimator(), "n_components")
```

aazuspan (changes requested):
This all looks great! I suggested one minor change, but otherwise (knock on wood) I think this should be ready to merge!

> I actually liked your naming suggestion for the function, so I kept it. I didn't inline it because we can use it for both tests (I think you meant to inlining the function within the test itself, right?)

That's what I meant, but wasn't thinking about the fact it would be duplicated. I agree with your choice.

> We can filter out the impossible combinations of (method, n_components), but in the test we also need to change the hyperpameters based on the estimator type as well. This means we're writing the same check (i.e. hasattr(estimator(), "n_components") both in the new estimator_does_not_support_n_components function as well as the test itself.

Yeah, no way to repurpose the `estimator_does_not_support_n_components` since it takes the weird result argument instead of an estimator. I'm okay with this, and maybe more importantly can't think of any way to improve it!

> In order to use estimator_does_not_support_n_components in both test_kneighbors and test_predict, the function signature has to accomodate result and n_components from test_kneighbors, but also the weighted parameter from test_predict. For now, the function signature looks like def estimator_does_not_support_n_components(result, n_components, **kwargs), but I'm not sure that's the right approach.

I see! That function signature is a little ugly, but I think your solution makes the most sense and it's pretty clear that it will just ignore any additional arguments.

tests/test_port.py
grovduck: I told you I was getting tired! 😉

aazuspan (approved):
LGTM!

PR #43, opened 2023-07-05
Migrate pre-commits to ruff

### LINKED ISSUE #41: Add ruff to pre-commit hooks
[pyupgrade](https://github.com/asottile/pyupgrade) standardizes syntax, including type annotations, to a minimum Python version. That should make it easier to maintain consistency, especially as we drop support for older versions of Python.

Good overview of the tool: [How to Upgrade Syntax with pyupgrade](https://adamj.eu/tech/2021/05/21/python-type-hints-how-to-upgrade-syntax-with-pyupgrade/)

EDIT: Rather than adding `pyupgrade`, we could switch to [ruff](https://github.com/astral-sh/ruff) which supports almost all of the features of `pyupgrade`, `flake8`, `isort`, and a variety of other linters (including `flake8` extensions like `bugbear` and `docstrings`.) In addition to being much faster than our current tools, this would consolidate almost everything into a single tool.

@grovduck let me know if you have any preferences here on tooling. Otherwise I'll probably try [migrating `sankee` to `ruff`](https://github.com/aazuspan/sankee/issues/50) over the weekend to get a better idea of whether there are any sticking points, and report back.

aazuspan:
This would close #41 by migrating most of the pre-commit checks to `ruff`. Specifically, `flake8` and `isort` are replaced and `pyupgrade` is added via `ruff`.

This also removes the disabled `sourcery` pre-commit hook per discussion in #40.

### Inconsistencies

There were a few small inconsistencies/surprises I encountered during the migration, detailed below:

1. When using `noqa` with a multiline string, e.g. a docstring that exceeds max line length, `ruff` wants the `noqa` at the end of the string instead of on the offending line. You can see that change in `_ccora.py`.
2. `ruff` (and `pyupgrade`) will only recommend updating type annotation syntax if `from __future__ import annotations` is already present in the file. I think this is [supposed to be fixed](https://beta.ruff.rs/docs/rules/future-rewritable-type-annotation/#future-rewritable-type-annotation-fa100) by combining the `pyupgrade` rules with the `flake8-future-annotations` rules, but for some reason I haven't been able to make that work (i.e. if you add a `Union` type to a file without the `__future__` import, `ruff` doesn't suggest any changes). I'm not sure if this is a bug in `ruff`, something wrong in my configuration, or a misunderstanding on my part. I'm going to keep looking into it.
3. Unlike `flake8`, I set up `ruff` to make fixes automatically if possible. I mostly did that for simplicity, but it is possible to set it up to [only fix some rules](https://beta.ruff.rs/docs/settings/#fixable), e.g. auto-fix import orders but only report linting issues, which would be more consistent with how things were set up previously. Any preference on this, @grovduck?

### Current rules

The current rules are defined below and [all supported rules are detailed here](https://beta.ruff.rs/docs/rules/):
https://github.com/lemma-osu/scikit-learn-knn-regression/blob/d410e81fe1bec868e8adb47795ab6efb35521a26/pyproject.toml#L61

I tried to include everything we already had, but I also included some other rules that seemed relevant and either passed with our existing code or caught minor issues.

@grovduck I'm opening this as a draft while I try and figure out `#2` above, and I'm also curious if you have any feedback or think there are any rules we should add/remove at this point.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,6 @@ include = [
 [tool.hatch.envs.default]
 dependencies = [
     "pre-commit",
-    # "sourcery",
 ]
 
 [tool.hatch.envs.test]
@@ -56,3 +55,12 @@ pythonpath = "src/"
 markers = [
     "uncollect_if(*, func): function to uncollect tests from parametrization",
 ]
+
+[tool.ruff]
+target-version = "py38"
+select = ["E", "I", "F", "B", "FA", "UP", "ISC", "PT", "NPY", "Q", "RET", "SIM", "PERF"]
+fix = true
+show-fixes = true
+
+[tool.ruff.isort]
+known-first-party = ["sknnr"]
--- a/src/sknnr/datasets/_base.py
+++ b/src/sknnr/datasets/_base.py
@@ -4,7 +4,7 @@ import csv
 import sys
 from dataclasses import dataclass
 from importlib import resources
-from typing import TYPE_CHECKING, TextIO, Union
+from typing import TYPE_CHECKING, TextIO
 
 import numpy as np
 from numpy.typing import NDArray
@@ -17,10 +17,10 @@ DATA_MODULE = "sknnr.datasets.data"
 
 @dataclass
 class Dataset:
-    index: Union[NDArray[np.int64], pd.DataFrame]
-    data: Union[NDArray[np.float64], pd.DataFrame]
-    target: Union[NDArray[np.float64], pd.DataFrame]
-    frame: Union[None, pd.DataFrame]
+    index: NDArray[np.int64] | pd.DataFrame
+    data: NDArray[np.float64] | pd.DataFrame
+    target: NDArray[np.float64] | pd.DataFrame
+    frame: None | pd.DataFrame
     feature_names: list[str]
     target_names: list[str]
 
@@ -90,7 +90,7 @@ def _load_csv_data(
 
 def load_moscow_stjoes(
     return_X_y: bool = False, as_frame: bool = False
-) -> Union[tuple[NDArray[np.float64], NDArray[np.float64]], Dataset]:
+) -> tuple[NDArray[np.float64], NDArray[np.float64]] | Dataset:
     """Load the Moscow Mountain / St. Joe's dataset (Hudak 2010[^1]).
 
     The dataset contains 165 plots with environmental, LiDAR, and forest structure
--- a/src/sknnr/transformers/_ccora.py
+++ b/src/sknnr/transformers/_ccora.py
@@ -4,8 +4,8 @@ from scipy.stats import f
 
 def ftest_cor(p, q, N, cor):
     """Direct port of yaImpute ftest_cor in yai
-    REFERENCE: https://github.com/jeffreyevans/yaImpute/blob/f06d1daf13cf461730c3da5f18f022c1074e9826/R/yai.R#L11-L42 # noqa: E501
-    """
+    REFERENCE: https://github.com/jeffreyevans/yaImpute/blob/f06d1daf13cf461730c3da5f18f022c1074e9826/R/yai.R#L11-L42
+    """  # noqa: E501
     s = min(p, q)
     k = np.arange(s) + 1
     L = np.array([np.prod(1 - np.square(cor[i:s])) for i in range(s)])
@@ -49,9 +49,9 @@ class CCorA:
 
     We were not able to use the statsmodels implementation directly as it detects
     collinear features and raises a `ValueError` which yaImpute allows to pass silently.
-    REFERENCE: https://github.com/statsmodels/statsmodels/blob/77cb066320391ffed4196a32491ddca28e8c9122/statsmodels/multivariate/cancorr.py#L17-L93 # noqa: E501
-    REFERENCE: https://github.com/jeffreyevans/yaImpute/blob/f06d1daf13cf461730c3da5f18f022c1074e9826/R/yai.R#L367-L371 # noqa: E501
-    """
+    REFERENCE: https://github.com/statsmodels/statsmodels/blob/77cb066320391ffed4196a32491ddca28e8c9122/statsmodels/multivariate/cancorr.py#L17-L93
+    REFERENCE: https://github.com/jeffreyevans/yaImpute/blob/f06d1daf13cf461730c3da5f18f022c1074e9826/R/yai.R#L367-L371
+    """  # noqa: E501
 
     TOLERANCE = 1e-8
     P_VAL = 0.05
--- a/tests/test_datasets.py
+++ b/tests/test_datasets.py
@@ -63,9 +63,10 @@ def test_load_moscow_stjoes_as_xy_as_frame():
 
 def test_asframe_raises_without_pandas():
     """Test that as_frame=True raises a helpful error if pandas is not installed."""
-    with mock.patch.dict(sys.modules, {"pandas": None}):
-        with pytest.raises(ImportError, match="pip install pandas"):
-            load_moscow_stjoes(as_frame=True)
+    with mock.patch.dict(sys.modules, {"pandas": None}), pytest.raises(
+        ImportError, match="pip install pandas"
+    ):
+        load_moscow_stjoes(as_frame=True)
 
 
 def test_dataset_repr():
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -40,9 +40,5 @@ jobs:
           python -m pip install --upgrade pip
           pip install hatch
 
-      # - name: Setup environment
-      #   run: |
-      #     hatch run sourcery login --token ${{ secrets.SOURCERY_TOKEN }}
-
-      - name: Set up pre-commit
+      - name: Run pre-commit hooks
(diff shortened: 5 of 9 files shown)


pyproject.toml
```diff
@@ -56,3 +55,12 @@ pythonpath = "src/"
 markers = [
     "uncollect_if(*, func): function to uncollect tests from parametrization",
 ]
+
+[tool.ruff]
+target-version = "py38"
+select = ["E", "I", "F", "B", "FA", "UP", "ISC", "PT", "NPY", "Q", "RET", "SIM", "PERF"]
```
grovduck, line 61: Will this configuration skip over tests like ['W'](https://beta.ruff.rs/docs/rules/#warning-w)?  Is there any redundancy across suites of tests such that issues that would be flagged on 'W' may be flagged elsewhere as well?  (EDIT: I think black will take care of 'W' rules, perhaps?)

I just tried to see what would pop out if we included *all* rules using this configuration:
```
[tool.ruff]
target-version = "py38"
ignore = []
show-fixes = true
```
but nothing came up with `precommit run --all-files`.  Is that the correct configuration and it just didn't find any issues?  If so, good on us 😄 

I always have the fear of missing out on some automated check, but those can quickly lead to annoyances, so I'm satified that you've probably captured the meaningful ones.

src/sknnr/datasets/_base.py
```diff
@@ -4,7 +4,7 @@
 import sys
 from dataclasses import dataclass
 from importlib import resources
-from typing import TYPE_CHECKING, TextIO, Union
+from typing import TYPE_CHECKING, TextIO
```
grovduck, line 7: Generic comment/question (not just this line).  I assume we'll need to use `from __future__ import annotations` on each module that we want to use the newer typing syntax, correct?  My brain wants to have this statement in one module and then import that module where we use it, but I don't think that saves us any effort and may actually add to our load.

src/sknnr/transformers/_ccora.py
```diff
@@ -4,8 +4,8 @@
 
 def ftest_cor(p, q, N, cor):
     """Direct port of yaImpute ftest_cor in yai
-    REFERENCE: https://github.com/jeffreyevans/yaImpute/blob/f06d1daf13cf461730c3da5f18f022c1074e9826/R/yai.R#L11-L42 # noqa: E501
-    """
+    REFERENCE: https://github.com/jeffreyevans/yaImpute/blob/f06d1daf13cf461730c3da5f18f022c1074e9826/R/yai.R#L11-L42
+    """  # noqa: E501
```
grovduck, line 8: Interesting, but probably not a big deal, right?

tests/test_datasets.py
```diff
@@ -63,9 +63,10 @@ def test_load_moscow_stjoes_as_xy_as_frame():
 
 def test_asframe_raises_without_pandas():
     """Test that as_frame=True raises a helpful error if pandas is not installed."""
-    with mock.patch.dict(sys.modules, {"pandas": None}):
-        with pytest.raises(ImportError, match="pip install pandas"):
-            load_moscow_stjoes(as_frame=True)
+    with mock.patch.dict(sys.modules, {"pandas": None}), pytest.raises(
+        ImportError, match="pip install pandas"
+    ):
+        load_moscow_stjoes(as_frame=True)
```
grovduck, line 69: Whoa! I've never seen that syntax before on a `with` statement (multiple statements).  I assume they happen sequentially.

grovduck (approved):
Super cool, @aazuspan.  Man, ruff is faaaaast.  Minor comments throughout and I don't think I have any proposed changes.  

Sounds like you're still wrestling with ruff not identifying issues associated with "old-style" typing.  I did try to edit `transformers/_base.py` with a `from __future__ import annotations` and switching to `Ordination = CCA | CCorA` and it passed pre-commit.  Presumably it would still fail when pushing.  But I think we knew that was going to be a problem.

Thanks for your great work!

src/sknnr/transformers/_ccora.py
aazuspan: Yeah, that's what I thought. I prefer the other a little, but it's a pretty minor drawback.

tests/test_datasets.py
aazuspan: That was new to me too! Recommended by [SIM117](https://beta.ruff.rs/docs/rules/multiple-with-statements/). From [the Python docs](https://docs.python.org/3/reference/compound_stmts.html#with):

image

src/sknnr/datasets/_base.py
aazuspan: > I assume we'll need to use from __future__ import annotations on each module that we want to use the newer typing syntax, correct?

Yes. For typing with built-in types (e.g. `list`) we can do that automatically once we go to [3.9](https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections), but using `|` instead of `Union` won't happen until [3.10](https://peps.python.org/pep-0604/), and as we figured out before, forward references are still indefinite. I'm guessing you're right and we'll end up needing the `__future__` import in just about every module for the foreseeable future.

I also had the idea of trying to automatically include it in all our modules somehow (I think I dislike the import because it's never explicitly used), but I landed at the same spot as you. I don't think there's any good way to do that, and even if there was I think it would be too surprising to silently modify behavior.

pyproject.toml
aazuspan: > Will this configuration skip over tests like ['W'](https://beta.ruff.rs/docs/rules/#warning-w)? Is there any redundancy across suites of tests such that issues that would be flagged on 'W' may be flagged elsewhere as well? (EDIT: I think black will take care of 'W' rules, perhaps?)

You're right that this doesn't include the `W` ruleset. I think there are potentially redundancies (and even conflicts) between some rulesets, but as you suspected my thought was that these specific cases should all be covered by `black`.

> I just tried to see what would pop out if we included all rules using this configuration: ... but nothing came up with precommit run --all-files. Is that the correct configuration and it just didn't find any issues? If so, good on us 😄

Sadly not! I think `ignore` only excludes selected rules, which defaults to just `E` and `F`. If you try again with `select = ["ALL"]` the results aren't quite as encouraging, although most of the things it catches aren't unexpected at least (docstrings, type annotations, and PEP8 naming are the big ones).

> I always have the fear of missing out on some automated check, but those can quickly lead to annoyances, so I'm satified that you've probably captured the meaningful ones.

It's easy enough to remove rules later if any turn out to be annoying, so I'm open to adding anything that looks useful to you! But it's also easy to add more later, so no need for us to finalize the list now.

aazuspan:
> Man, ruff is faaaaast.

Right?! I know `sknnr` isn't big enough for that to be a huge deal, but it's still pretty impressive.

> I did try to edit transformers/_base.py with a from __future__ import annotations and switching to Ordination = CCA | CCorA and it passed pre-commit. Presumably it would still fail when pushing. But I think we knew that was going to be a problem.

Yeah, I wish the `pyupgrade` rules would catch those, but alas it seems to just be focused on forward compatibility.

> Just found one minor issue. In README.md, the example you give uses pre-commit run flake8 which no longer exists. I'm OK taking out that section or changing it to one of the others (probably black or mypy).

Good catch, thanks! I'll just update it to a relevant tool for now, but I assume we'll move most of the current readme into a contributing guide eventually.

Thanks for the review! I'll poke around with the future annotation rules a little more, and let you know if I find anything or give up.

pyproject.toml
grovduck: I think now that I know the trick about how to run with "ALL" (thanks!), I can play around with it (outside of this PR) and see if there are any that I find useful.  It brought up a lot of stuff with `_cca.py` and `_ccora.py` that I will probably adhere to, but like you I'm OK not finalizing the list right now.

aazuspan:
Update: finally tracked down the `flake8-future-annotations` rule issue to what I'm *pretty sure* is a bug in `ruff`. See https://github.com/astral-sh/ruff/issues/5574.

We may want to wait a bit and see what happens over there, but I'm guessing it will make sense to just merge this as is and pin `ruff` later once a fix is released.

grovduck:
> Update: finally tracked down the `flake8-future-annotations` rule issue to what I'm _pretty sure_ is a bug in `ruff`

Sweet sleuthing and a quick fix to boot!

aazuspan:
> quick fix

No kidding! Given the release cadence over there (5 releases in the last 2 weeks 🤯), maybe we're better off just sitting tight and merging this once that's out.

aazuspan:
And [it's out](https://github.com/astral-sh/ruff/releases/tag/v0.0.278)! I bumped the pre-commit hook to the new `ruff` version, so it will now catch any outdated type annotations that could be simplified with a `__future__` import.

I was worried it would incorrectly flag the line below (which as we discovered cannot be fixed by the `__future__` import), but it looks like it's smarter than we are, so it's not a problem :stuck_out_tongue: 

https://github.com/lemma-osu/scikit-learn-knn-regression/blob/950daf6e5395dd1b2a7e1a4a3c3cf46cebb59e57/src/sknnr/transformers/_base.py#L11

I don't see any unresolved issues, so unless you've had any more thoughts on rules, I think I'm probably good to merge?

grovduck:
> but it looks like it's smarter than we are

Most things are smarter than I am!

Issue #42, opened 2023-07-04
Replace existing port tests with automated regression tests


aazuspan:
Currently, we test our estimator accuracy against manually generated results from `yaImpute` and `pynnmap`. Once we have all major functionality implemented and are confident in our results, we can switch to regression testing to ensure no errors are introduced. This was briefly discussed in #40:

> Eventually, I suppose we will reach a point where we're confident everything is working and could theoretically test against previous versions of sknnr ... [pytest-regressions](https://github.com/ESSS/pytest-regressions) and [syrupy](https://github.com/tophat/syrupy) both look like interesting tools that might solve this problem for us in the future by testing against automatically generated results.

We'll need to do some experimenting to find the right tool for our use case, but the advantage should be a massive simplification of our testing system (e.g. the removal of `KNNTestDataset`) and the ability to easily test against a wider range of parameters (e.g. values of `n_neighbors` and `n_components`).

grovduck:
I started experimenting a bit with `syrupy` and it seems like a pretty straightforward API.  It looks like we might need to put special care into how we serialize big arrays as the default in `syrupy` looks to store each element on a separate line in the resulting `.ambr` file.  It looks like `pytest-regressions` might have better support for the types of data that we will want to test (i.e. arrays and dataframes) or we could use JSON serialization together with `syrupy` to get the output a bit more compact.  I was playing around with `orjson` and this test:
```
import numpy as np
import orjson

def test_arr_1(snapshot):
    arr = np.arange(24).reshape(4, 6)
    assert orjson.dumps(arr, option=orjson.OPT_SERIALIZE_NUMPY) == snapshot
```
results in this serialization:
```
# name: test_arr_1
  b'[[0,1,2,3,4,5],[6,7,8,9,10,11],[12,13,14,15,16,17],[18,19,20,21,22,23]]'
# ---
```
whereas this test:
```
def test_arr_2(snapshot):
    arr = np.arange(24).reshape(4, 6)
    assert_array_equal(arr, snapshot)
```
results in this snipped serialization:
```
# name: test_arr_2
  0
# ---
# name: test_arr_2.1
  1
# ---
# name: test_arr_2.10
  10
# ---
<snip>
# name: test_arr_2.7
  7
# ---
# name: test_arr_2.8
  8
# ---
# name: test_arr_2.9
  9
# ---
```

aazuspan:
[...]

PR #46, opened 2023-08-29
Add R6 southwest Oregon (SWO) Ecology dataset

grovduck:
This PR would add the R6 southwest Oregon (SWO) Ecology plot dataset to the sample data available in `sknnr`.  The dataset consists of a species matrix (percent cover by tree species) and an environmental matrix (climate, topography, and spectral) for 3,005 plots in SWO measured in 2000.  The R6 Ecology Program installed these plots and this PR is contingent upon their permission to use these data.  

This PR adds a new function (`load_swo_ecoplot`) which makes these data available to all `sknnr` (and `scikit-learn`) estimators.

~~Pending approval, the data in `swo_ecoplot_env.csv` and `swo_ecoplot_spp.csv` are all dummy values, but do currently pass our test suite.~~ EDIT: Actual data is now committed after approval from R6 Ecology Group and data citation is added.


src/sknnr/datasets/_base.py
```diff
@@ -126,22 +173,55 @@ def load_moscow_stjoes(
...
+
+    Returns
+    -------
+    Dataset or tuple of ndarray
+        A Dataset object containing the data, target, and feature names. If return_X_y
+        is True, return a tuple of data and target arrays instead.
+
+    Notes
+    -----
+    TBD
```
grovduck, line 213: Obviously, I need to revisit these TBDs once we hear back.

tests/test_datasets.py
```diff
@@ -1,74 +1,108 @@
...
-    assert y.shape == (165, 35)
+    assert X.shape == (configuration.n_samples, configuration.n_features)
+    assert y.shape == (configuration.n_samples, configuration.n_targets)
     assert_array_equal(X.index.values, data.index)
     assert_array_equal(y.index.values, data.index)
 
 
-def test_asframe_raises_without_pandas():
+@pytest.mark.parametrize("configuration", CONFIGURATIONS, ids=CONFIGURATION_IDS)
+def test_asframe_raises_without_pandas(configuration: DatasetConfiguration):
```
grovduck, line 93: Perhaps not necessary to test both configurations on this one - I don't know why one would raise and the other wouldn't.  But I was doing blanket replacement and chose to do this one as well.

tests/test_datasets.py
```diff
@@ -1,74 +1,108 @@
...
 import pytest
 from numpy.testing import assert_array_equal
 
-from sknnr.datasets import load_moscow_stjoes
+from sknnr.datasets import load_moscow_stjoes, load_swo_ecoplot
 
 
-def test_load_moscow_stjoes():
+@dataclass
+class DatasetConfiguration:
```
aazuspan, line 14: Nice clean solution!

tests/test_datasets.py
```diff
@@ -1,74 +1,108 @@
...
+CONFIGURATIONS = [
+    DatasetConfiguration(
+        load_function=load_moscow_stjoes, n_samples=165, n_targets=35, n_features=28
+    ),
+    DatasetConfiguration(
+        load_function=load_swo_ecoplot, n_samples=3005, n_targets=25, n_features=18
+    ),
+]
+
+CONFIGURATION_IDS = ["moscow_stjoes", "swo_ecoplot"]
```
aazuspan, line 30: Can we combine these into a dict with the IDs as keys and configs as values, just to ensure they stay synced?

tests/test_datasets.py
aazuspan: Yeah, not necessary but it doesn't do any harm. I'm good with that choice.

src/sknnr/datasets/_base.py
```diff
@@ -88,6 +88,53 @@ def _load_csv_data(
     return index, data, data_names
 
 
+def _load_dataset_from_filenames(
```
aazuspan, line 91: What do you think about `_load_dataset_from_csvs`, just to be a little more specific? I don't have a strong opinion here, so whatever you prefer works.

aazuspan (commented):
This looks great! I think generalizing the data loading and testing was a good choice, especially if we end up adding other datasets down the line. Much cleaner than the sklearn approach.

My comments are all very minor.

src/sknnr/datasets/_base.py
```diff
@@ -88,6 +88,53 @@ def _load_csv_data(
...
+    *,
+    data_filename: str,
+    target_filename: str,
+    return_X_y: bool = False,
+    as_frame: bool = False,
+) -> tuple[NDArray[np.float64], NDArray[np.float64]] | Dataset:
+    """Generic loader for datasets with separate data and target files.
```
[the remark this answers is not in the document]
grovduck: Good idea!  I copied/pasted the docstring from the Notes section of `_load_csv_data` and added a sentence to note that both datasets' IDs should match.  Please add more detail if you think it needs it.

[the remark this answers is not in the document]
grovduck: Yep, you're right.  I've edited this, but not sure it's quite right.  Suggestions definitely welcome.

tests/test_datasets.py
grovduck: Perfect.  I've taken a stab at this.

src/sknnr/datasets/_base.py
grovduck: I tried to marry both our suggestions.  I went with `_load_dataset_from_csv_filenames`.  A bit long, but my brain wasn't reading `csvs` as the plural of `csv` - more like "cis-vis" :smile:.

aazuspan: That's fair, looks good to me!

grovduck:
Just FYI, I ran these data through yaImpute and created a bunch of new train/test files.  Then I clumsily copied `test_port.py` into a `swo_ecoplot` specific version and all tests passed.  I didn't think it was worth adding all of these files just to verify the porting logic, but we may want to keep the yaImpute output around if we end up getting back to #42?

aazuspan:
Definitely reassuring to hear that everything passes with a second dataset, but I agree that it's not needed in the test suite. For the connection with #42, are you thinking that we would verify accuracy against yaImpute one last time before we switch to regression testing and ditch the port tests? That sounds reasonable to me!

grovduck:
> For the connection with #42, are you thinking that we would verify accuracy against yaImpute one last time before we switch to regression testing and ditch the port tests? That sounds reasonable to me!

I'm not sure what I was thinking, but I did just spend a bit of time playing around with `syrupy` and I think I understand the main concepts there.  I'll put some follow up thoughts in #42.

grovduck:
@aazuspan.  I think this is ready to go, but thought I'd see if you had any issues with the docstring for `load_swo_ecoplot`.

aazuspan:
LGTM!

Issue #47, opened 2023-09-07
Using properties as accessors in CCA/CCorA leads to slow run times


grovduck:
In early experiments with putting a spatial wrapper around sknnr estimators (and specifically `GNNRegressor`) using [sknrr-spatial](https://github.com/lemma-osu/sknnr-spatial), @aazuspan noticed that the performance of `CCATransformer.transform` was not covarying with `KNeighborsRegressors.kneighbors` as expected with different chunk sizes.  Smaller chunk sizes would lead to `CCATransformer.transform` actually taking a majority of the CPU time as profiled with [pyinstrument](https://github.com/joerick/pyinstrument).  We're capturing the conversation here to decide on next steps to fix this issue.

grovduck:
From @aazuspan:

Just a quick follow up after running a little bit of performance profiling with pyinstrument on `GNNRegressor.kneighbors`: The proportion of CPU time spent on `KNeighborsRegressor.kneighbors` vs. `GNNRegressor._apply_transform` varies with the size of X. With 4096x4096 samples, `_apply_transform` should only take up about 8.5% of the time.

![image](https://github.com/lemma-osu/scikit-learn-knn-regression/assets/1706997/fa6d0b62-127f-49e7-b1a1-b4beed620a8d)
 
On the other hand, if the dask chunks were really small, that proportion might shift to where `_apply_transform` takes the majority of the time. Here's the result with 64x64 samples, where `_apply_transform` takes 86% of the CPU time for kneighbors.

![image](https://github.com/lemma-osu/scikit-learn-knn-regression/assets/1706997/68726d7e-78b9-42b1-93f6-b4274bdb06cb)

Maybe there's a substantial overhead to the `_apply_transform` method that makes it relatively inefficient for small datasets?

grovduck:
From @grovduck:

Seeing your results is really surprising to me.  I varied chunk sizes from 32x32 up to 4096x4096.  I’m completely flummoxed why these two methods (`_apply_transform` and `kneighbors`) wouldn’t scale linearly relative to one another, independent of chunk size.  Do you have any ideas off the top of your head why chunk size would affect their relative performance?

grovduck:
From @aazuspan:

(@aazuspan provided a Jupyter notebook with basic profiling code to replicate the issue):

```
[cell]
%load_ext pyinstrument
%load_ext autoreload
%autoreload 2

[cell]
from sknnr.datasets import load_swo_ecoplot
[...]

PR #48, opened 2023-09-11
Store repeatedly used variables as estimator attributes #47

grovduck:
For `CCATransformer` and `CCorATransformer`, store matrices from ordination as estimator attributes within `fit` method.  Previously, we were calling properties from `CCA` and `CCorA` (respectively) that were recomputing expensive operations.  Because the matrix projectors don't change once fit, it's more efficient to store as estimator attributes.

Also, remove redundant call to `_validate_data` in `TransformedKNeighborsMixin._apply_transform` as the transformers are now responsible for that call.

### DIFF
--- a/src/sknnr/_base.py
+++ b/src/sknnr/_base.py
@@ -70,7 +70,6 @@ class TransformedKNeighborsMixin:
     def _apply_transform(self, X) -> NDArray:
         """Apply the stored transform to the input data."""
         check_is_fitted(self, "transform_")
-        self.transform_._validate_data(X, reset=False)
         return self.transform_.transform(X)
 
     def fit(self, X, y):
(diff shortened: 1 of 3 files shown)


aazuspan (approved):
LGTM!

I know there are still repeated property calls in the underlying transformer implementations (e.g. the `ConstrainedOrdination.env_center` property gets recalculated multiple times) that could be simplified with the same strategy as this and probably improve performance further. I'd say those *could* be included in this PR, but getting this merged quick to keep things moving sounds reasonable to me.

grovduck:
> I know there are still repeated property calls in the underlying transformer implementations (e.g. the `ConstrainedOrdination.env_center` property gets recalculated multiple times) that could be simplified with the same strategy as this and probably improve performance further. I'd say those _could_ be included in this PR, but getting this merged quick to keep things moving sounds reasonable to me.

I absolutely agree that there is more work to be done on `CCA` and `CCorA`.  I'd like to defer that for now, but have created #49 to track this.  Merging now to keep things moving seems like a good tack to me.

PR #52, opened 2023-09-18
Refactor transformed estimators

### LINKED ISSUE #51: Refactor transformed estimators
@grovduck I have a proposal to run by you, inspired by your refactor of the inheritance in #50. This design builds off of the changes there, so that would be merged before tackling this. I thought about proposing this as part of that PR, but didn't want to derail things with even more refactoring.

I suggest we turn `TransformedKNeighborsRegressor` into an abstract class by inheriting from [ABC](https://docs.python.org/3/library/abc.html#abc.ABC), then add an abstract method `_get_transform` that all subclasses would be required to implement, e.g.:

```python
class EuclideanKNNRegressor(TransformedKNeighborsRegressor):
    def _get_transform(self) -> TransformerMixin:
        return StandardScalerWithDOF(ddof=1)

```

Next, we would get rid of the `fit` methods on those estimators and move that functionality into `TransformedKNeighborsRegressor.fit`, using a `_set_fit_transform` method to handle the instantiation and fitting of the transformer.

```python
class TransformedKNeighborsRegressor(RawKNNRegressor, ABC):
    ...

    @abstractmethod
    def _get_transform(self) -> TransformerMixin:
        """Return the transformer to use for fitting. Must be implemented by subclasses."""
        ...

    def _set_fit_transform(self, X, y) -> TransformerMixin:
        self.transform_ = self._get_transform().fit(X, y)

    def fit(self, X, y):
        """Fit using transformed feature data."""
        self._validate_data(X, y, force_all_finite=True, multi_output=True)        
        self._set_dataframe_index_in(X)
        self._set_fit_transform(X, y)

        X_transformed = self.transform_.transform(X)
        return super().fit(X_transformed, y)

```

To accommodate for fitting transformers with separate y data in `GNN` and `MSN`, we could add a kNN-independent `YFitMixin` that overrides the `fit` method to accept the additional argument, store it as an attribute, and fit the transformer with it using an overriden `_set_fit_transform`:

```python
class YFitMixin:
    """Mixin for transformed estimators that can fit the transformer with a separate y."""
    def _set_fit_transform(self, X, y):
        y_fit = self.y_fit_ if self.y_fit_ is not None else y
        self.transform_ = self._get_transform().fit(X, y_fit)

    def fit(self, X, y, y_fit=None):
        self.y_fit_ = y_fit
        return super().fit(X, y)
```

Overall, this should reduce some code duplication in the `fit` methods, prevent instantiation of `TransformedKNeighborsRegressor` without having to rely on making it private, and add a runtime check to ensure that all transformed estimators define a transformer function. The main downside is the need to store a reference to `y_fit_`. There may be another way to handle the `YFitMixin`, but that was the best solution I could come up with after trying a few different strategies.

Curious to hear your thoughts on this design, and if you foresee any limitations.

aazuspan:
This would resolve #51 by: 

- Refactoring the `TransformedKNeighborsRegressor` into an abstract class
- Moving transformer instantiation out of the estimator `fit` methods and into an abstract `_get_transformer` method
- Moving transformer fitting out of the estimator `fit` methods and into the `TransformedKNeighborsRegressor._set_fitted_transformer` method to reduce duplication
- Creating a `YFitMixin` to handle transformer fitting for `GNN` and `MSN`

This also:
- Renames the `transform_` attribute to `transformer_` for consistency with the new methods
- Adds a `_validate_data` check with `force_all_finite=True` when fitting all transformed estimators. This was needed by `MSN`, but also fixed an xfailing estimator check for `GNN`, which allowed us to drop that from the tags.

@grovduck Sorry about the convoluted commit history... Not sure exactly how, but I accidentally included all of your commits from #50 in this. Once this gets squashed it shouldn't be a problem at least.


src/sknnr/_gnn.py
```diff
@@ -1,16 +1,14 @@
...
+class GNNRegressor(YFitMixin, TransformedKNeighborsRegressor):
     def __init__(self, n_components=None, **kwargs):
         super().__init__(**kwargs)
         self.n_components = n_components
 
-    def fit(self, X, y, y_fit=None):
-        y_fit = y_fit if y_fit is not None else y
-        self.transform_ = CCATransformer(self.n_components).fit(X, y=y_fit)
-        return super().fit(X, y)
+    def _get_transformer(self):
```
grovduck, line 10: Should we be providing type information here as you did for `EuclideanKNNRegressor` (i.e. `TransformerMixin`).  Also on `MahalanobisKNNRegressor` and `MSNRegressor`

grovduck (approved):
@aazuspan, all looks great to me.  Exactly as you laid it out ...

src/sknnr/_gnn.py
aazuspan: Definitely, good call!

aazuspan:
Thanks @grovduck, merging!

PR #54, opened 2023-09-20
Test `y_fit` functionality

### LINKED ISSUE #53: Test estimators with `y_fit` data
While working on #52, I noticed that we don't currently have any tests to confirm the functionality or accuracy of predictions that use separate `y_fit` data. This caused a few bugs (e.g. https://github.com/lemma-osu/scikit-learn-knn-regression/pull/52/commits/7f92a90e62a2da640e920dde418b37b3b226002d) to slip by since the tests were passing even if `y_fit` was ignored.

@grovduck, I'm not 100% sure we need a full-blown port test for this. We already know that we can accurately fit transformers, so maybe all we need to do is confirm that they're getting fit with the correct data? I know it's a pain to add in a whole new set of test data (and I'm not even sure if this is doable with `yaImpute`?), so I'm open to an easy solution, as long as we're confident that it can work as expected. What do you think?

aazuspan:
Resolves #53 by testing functionality of the `y_fit` parameter for `MSN` and `GNN` estimators.

### DIFF
--- a/tests/test_estimators.py
+++ b/tests/test_estimators.py
@@ -22,6 +22,8 @@ TEST_ESTIMATORS = [
     GNNRegressor,
 ]
 
+TEST_YFIT_ESTIMATORS = [MSNRegressor, GNNRegressor]
+
 
 @pytest.mark.filterwarnings("ignore:divide by zero encountered")
 @parametrize_with_checks([cls() for cls in TEST_ESTIMATORS])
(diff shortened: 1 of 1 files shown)


tests/test_estimators.py
```diff
@@ -118,3 +120,33 @@ def test_estimator_output_type_consistency(output_mode, x_type, estimator):
         ref_type = type(ref_estimator.fit(X, y).predict(X))
 
     assert sknnr_type is ref_type  # noqa: E721
+
+
+@pytest.mark.parametrize("estimator", TEST_YFIT_ESTIMATORS)
+def test_yfit_is_stored(estimator):
+    """Test that y_fit is stored when passed."""
+    X, y = load_moscow_stjoes(return_X_y=True)
+    # Arbitrary split with a fixed constant to prevent zero sum rows
```
aazuspan, line 129: Too hacky? I couldn't easily find a combo of columns that prevented zero sum row errors, but I also didn't spend a ton of time looking...

tests/test_estimators.py
```diff
@@ -118,3 +120,33 @@ def test_estimator_output_type_consistency(output_mode, x_type, estimator):
...
+    assert_array_equal(est.y_fit_, y_fit)
+
+
+@pytest.mark.parametrize("estimator", TEST_YFIT_ESTIMATORS)
+def test_yfit_affects_prediction(estimator):
+    """Test that y_fit affects predictions when passed."""
+    X, y = load_moscow_stjoes(return_X_y=True)
+    # Arbitrary split with a fixed constant to prevent zero sum rows
+    y_fit = y[:, 10:] + 0.1
+    y = y[:, :10] + 0.1
```
aazuspan, line 145: I don't love the duplication, but also didn't want to test multiple things at once. Let me know if you think there's a better approach.

tests/test_estimators.py
grovduck: Not at all.  I think this is totally fine, especially with the comment.

tests/test_estimators.py
grovduck: I suppose you could put the data into a fixture and use that, e.g.
```
@pytest.fixture
def fit_data():
    X, y = load_moscow_stjoes(return_X_y=True)
    # Arbitrary split with a fixed constant to prevent zero sum rows
    y_fit = y[:, 10:] + 0.1
    y = y[:, :10] + 0.1
    return X, y, y_fit

...

@pytest.mark.parametrize("estimator", TEST_YFIT_ESTIMATORS)
def test_yfit_is_stored(estimator, fit_data):
    """Test that y_fit is stored when passed."""
    X, y, y_fit = fit_data
    ...

@pytest.mark.parametrize("estimator", TEST_YFIT_ESTIMATORS)
def test_yfit_affects_prediction(estimator, fit_data):
    """Test that y_fit affects predictions when passed."""
    X, y, y_fit = fit_data
   ...
```
but not sure that helps the readibility.  I'm fine with the duplication.

grovduck (approved):
@aazuspan, I think this is nice and clean and I don't think it's necessary to test with real data.  If I can't help myself and decide to get some results based on separate `y` and `y_fit` from `yaImpute`, I don't think there is any need to bring that into tests.  It would be a one-off for me.

tests/test_estimators.py
aazuspan: I think a fixture is a good idea!

grovduck:
LGTM!

PR #56, opened 2023-09-21
Only store array-like `dataframe_index_in_` and fix `CCA` compability with `list`

### LINKED ISSUE #55: `list.index` accidentally stored as `dataframe_index_in_`
I just came across scikit-learn/scikit-learn#27037 that points out that using `hasattr` or `getattr` to retrieve a DataFrame `index` will unintentionally retrieve the `list.index` method. For us, this means that an estimator fitted with a `list` will store that method as `dataframe_index_in_`. Then, if you used `return_dataframe_index=True` with `kneighbors`, you'd get a weird `IndexError`. 

```python
from sknnr import EuclideanKNNRegressor
from sknnr.datasets import load_swo_ecoplot

X, y = load_swo_ecoplot(return_X_y=True)
X = X.tolist()

est = EuclideanKNNRegressor().fit(X, y)
print(est.dataframe_index_in_)  # <built-in method index of list object at 0x000001E8BB330BC0>
est.kneighbors(return_dataframe_index=True)  # IndexError: too many indices for array
```

This is the offending line:

https://github.com/lemma-osu/scikit-learn-knn-regression/blob/96a4230cbbd113136aa619f869afabe1d648e054/src/sknnr/_base.py#L14-L15

They solve this in scikit-learn/scikit-learn#27044 by using a new [`_is_pandas_df`](https://github.com/thomasjpfan/scikit-learn/blob/24c7d17d57747e84be76ae1d0ddf12d82aaeeaa5/sklearn/utils/validation.py#L1990) helper, but as far I can tell that's not released yet, so I think it'll be a while before we want to rely on that. 

We could add a similar function to `sknnr`, but I wonder if we might break some implicit duck-typed support for `DataFrame`-like objects by adding an explicit class check here. An alternative would just be to check that `index` is an array using [`_is_arraylike`](https://github.com/scikit-learn/scikit-learn/blob/ba7d86956da03aa4fd230b1bbe8df57b63cf2dc0/sklearn/utils/validation.py#L268), which would prevent this bug and should cover us for just about any other case.

aazuspan:
This resolves #55 by only storing the `dataframe_index_in_` if it's array-like, and updates `test_estimators_support_dataframe_indexes` to confirm that the `list.index` method isn't accidentally stored like it previously was.

While testing this I noticed a small bug that prevented fitting `CCATransformer` with a `y: list` due to the dimensionality check. Just changing the order of casting and checking shape solved that, so I included it here as a quick fix.

### DIFF
--- a/src/sknnr/transformers/_cca_transformer.py
+++ b/src/sknnr/transformers/_cca_transformer.py
@@ -16,11 +16,10 @@ class CCATransformer(ComponentReducerMixin, TransformerMixin, BaseEstimator):
             ensure_min_features=2,
             ensure_min_samples=1,
         )
-
-        if not hasattr(y, "shape") or len(y.shape) < 2:
+        y = np.asarray(y)
+        if len(y.shape) < 2:
             raise ValueError("`y` must be a 2D array.")
 
-        y = np.asarray(y)
         self.ordination_ = CCA(X, y)
         self.set_n_components()
         self.env_center_ = self.ordination_.env_center
--- a/tests/test_estimators.py
+++ b/tests/test_estimators.py
@@ -80,6 +80,10 @@ def test_estimators_support_dataframe_indexes(estimator):
     with pytest.raises(NotFittedError, match="fitted with a dataframe"):
         estimator.kneighbors(return_dataframe_index=True)
 
+    # Make sure that `list.index()` is not accidentally stored
+    estimator.fit(moscow.data.tolist(), moscow.target)
+    assert not hasattr(estimator, "dataframe_index_in_")
+
     estimator.fit(X_df, moscow.target)
     assert_array_equal(estimator.dataframe_index_in_, moscow.index)
 
--- a/tests/test_estimators.py
+++ b/tests/test_estimators.py
@@ -93,6 +93,14 @@ def test_estimators_support_dataframe_indexes(estimator):
 
 
 @pytest.mark.parametrize("estimator", TEST_ESTIMATORS)
+def test_estimators_support_lists(estimator):
+    """All estimators should fit and predict data stored as lists."""
+    X, y = load_moscow_stjoes(return_X_y=True)
+    estimator = estimator().fit(X.tolist(), y.tolist())
+    estimator.predict(X.tolist())
+
+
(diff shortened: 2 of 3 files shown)


src/sknnr/transformers/_cca_transformer.py
```diff
@@ -16,11 +16,10 @@ def fit(self, X, y):
             ensure_min_features=2,
             ensure_min_samples=1,
         )
-
-        if not hasattr(y, "shape") or len(y.shape) < 2:
+        y = np.asarray(y)
+        if len(y.shape) < 2:
             raise ValueError("`y` must be a 2D array.")
 
-        y = np.asarray(y)
```
grovduck, line 23: Good catch!  Makes total sense.

tests/test_estimators.py
```diff
@@ -80,6 +80,10 @@ def test_estimators_support_dataframe_indexes(estimator):
     with pytest.raises(NotFittedError, match="fitted with a dataframe"):
         estimator.kneighbors(return_dataframe_index=True)
 
+    # Make sure that `list.index()` is not accidentally stored
+    estimator.fit(moscow.data.tolist(), moscow.target)
+    assert not hasattr(estimator, "dataframe_index_in_")
+
```
grovduck, line 86: Excellent test!

tests/test_estimators.py
```diff
@@ -88,6 +92,14 @@ def test_estimators_support_dataframe_indexes(estimator):
     assert_array_equal(idx.ravel(), moscow.index)
 
 
+@pytest.mark.parametrize("estimator", TEST_ESTIMATORS)
+def test_estimators_support_lists(estimator):
+    """All estimators should fit and predict data stored as lists."""
+    X, y = load_moscow_stjoes(return_X_y=True)
+    estimator = estimator().fit(X.tolist(), y.tolist())
+    estimator.predict(X.tolist())
```
grovduck, line 100: This is nice assurance as well.  I assume this is where you found the `CCA` bug.

grovduck (approved):
@aazuspan, I don't see anything that gives me pause here.  Looks good to me if you're ready to merge!

tests/test_estimators.py
aazuspan: Exactly!

PR #57, opened 2023-09-21
Core estimators

aazuspan:
This would close out the [Core estimators milestone](https://github.com/lemma-osu/scikit-learn-knn-regression/milestone/2) by merging our development branch into `main` :tada: 

@grovduck I'd propose we do a merge commit to preserve all the commits that went into the branch, even though there are a few messier ones early on before we started squashing. EDIT: Alternatively I guess we could do some rebasing to clean up the earlier commits, but I'm not sure that's really necessary...


grovduck:
Woohoo! :tada:

> I'd propose we do a merge commit to preserve all the commits that went into the branch, even though there are a few messier ones early on before we started squashing. EDIT: Alternatively I guess we could do some rebasing to clean up the earlier commits, but I'm not sure that's really necessary...

Absolutely I agree with doing a merge commit.  I'm OK retaining the messy early commits as proof of our (my!) growth over the history of this milestone, but if it bothers you, I'm OK rebasing as well.  Not sure anyone will care but us!

grovduck:
I'll have you do the honors when you're ready!  🤞

aazuspan:
It is done! 46 commits and ~12k lines (admittedly most of those are CSVs, but still). Not bad for 5 months of work!

grovduck:
> Not bad for 5 months of work!

Crazy!  Great way to end the week.

PR #59, opened 2023-09-29
Set up documentation

aazuspan:
This would close #5 by setting up documentation for `sknnr` using `mkdocs`.

@grovduck This might be a little premature, but I was excited to try out `mkdocs` (big fan so far!), and I figured having the docs in place will make things easier once we start digging into #10. I'm considering this an early prototype, as there's still tons of room to adjust the layout, visuals, and (especially) content.

I moved the developer guide into `docs/pages/contributing.md` and added some documentation-specific info there. The most important thing is to run `hatch run docs:serve` to start a live-reloading server where you can preview the docs.

### Guides

This uses [mkdocs](https://www.mkdocs.org/) for the core functionality, [mkdocstrings](https://mkdocstrings.github.io/) to generate the API reference from docstrings, and [mkdocs-material](https://squidfunk.github.io/mkdocs-material/getting-started/) for the theme, which adds a lot of nice additional functionality. The [mkdocstrings guide](https://mkdocstrings.github.io/python/usage/) details the features and some of the config options for generating the API reference. I suspect once we start getting more complete docstrings, it may become more apparent how we want to set that up. The [mkdocs-material config guide](https://squidfunk.github.io/mkdocs-material/creating-your-site/#advanced-configuration) has a lot of info on customizing the site appearance and functionality. The [mkdocs config guide](https://www.mkdocs.org/user-guide/configuration/#formatting-options) also has some good info on markdown extensions and plugins that might be of interest. I definitely haven't done a comprehensive read through any of those guides, so if you find anything cool that's worth adding, feel free to make changes!

### Content

I wanted to rough out the content just to give us a starting point, but as the domain expert I'm hoping you can rewrite and flesh things out as needed. Feel free to modify anything I wrote! My assumption is that this is being written for an audience that's already familiar with `sklearn` so we don't necessarily need to dive into the technical basics, but I'm not as sure about how much detail on the stats to include. I don't think it would hurt to discuss how the estimators work, but we could also just point to the relevant papers.

### ReadTheDocs

You can see the RTD build for this PR [here](https://sknnr--59.org.readthedocs.build/en/59/), and it should update whenever we push a new commit. Once this gets merged into main, RTD will build that and we'll have a stable docs page. You can track build status [here](https://readthedocs.org/projects/sknnr/builds/). Currently I'm the RTD maintainer for the project, but if you want to create an RTD account (I just login with my Github), I can add you as a maintainer, which would allow you to manually trigger builds and access project settings.

PR #60, opened 2023-10-02
WIP: Refactor ordination classes to remove extraneous details

### LINKED ISSUE #49: Refactor `CCA` and `CCorA` classes to reduce repeated operations
At present, `CCA` and `CCorA` are implemented as classes that mainly use `@property` to access attributes.   As such, properties call other properties within the class and there are often cases when properties are used repeatedly, which forces recalculation.  This is particularly expensive when evaluating methods from the `np.linalg` module such as singular value decomposition, QR decomposition, and solving for systems of linear equations.  We need to ensure that attributes are only calculated once when there is no possibility of change to attribute values.

This issue is closely tied to #47, but we are opening it in a separate issue for a few different reasons:

1. The `transform` method on the enclosing estimators (`CCATransformer` and `CCorATransformer`) is the method that is called most often and the fixes in #48 address the issue of repeated calls to properties by setting estimator attributes during `fit`.  Because the `fit` method on these transformers calls the classes of interest in this issue (`CCA` and `CCorA`), it still suffers from repeated calls to properties, but `fit` is typically called only once, so delaying the refactor shouldn't be too problematic.
2. Initially, when we first wrote the code for these classes, we were mostly concerned with the correct porting behavior from `yaImpute` and wanted to verify that "checkpoints" in the porting process were lining up with output from yaImpute.  As such, there are properties that correspond to multiple steps in each ordination process, very few of which we actually use.  Now that the porting seems to be behaving correctly, we can treat many of these properties as local variables and only retain the attributes/properties/methods needed for clients and enclosing classes.  However, if there is value in retaining some of these "internal" attributes for diagnosing model goodness-of-fit (e.g. charting capabilities for ordinations), we should perhaps retain these as (at the very least) private properties/methods.  This will require a bit of research into what attributes are retained for diagnostics, using `yaImpute` and `vegan` for inspiration.
3. For at least `CCA`, there are actually a number of other ordination techniques (redundancy analysis (RDA), distance-based RDA) that use common code already implemented in this class from which we may choose to create additional estimators.  Before optimizing the two existing classes, we'll want to think through the design of any new estimators and reuse as much of the existing code as possible.

Note that none of these reasons really stops us from refactoring out the duplicate calls at present, but changes to the existing code and potential expansion into a family of classes makes it prudent to delay until after the `Core Estimators` milestone.

grovduck:
This PR will close #49 once completed.  At present, it only addresses `CCA` and will require further changes based on class designs that have changed.  Here are the current changes and some comments which need to be addressed before proceeding.
 
#### Changes
- Remove all unreferenced properties in `OrdConstrained` and renamed to `ConstrainedOrdination`
- Change properties to variables when property is not called outside class
- Change `ContrainedOrdination` to abstract class with two abstract methods: `_transform_X` and `_transform_Y`.
- Refactored `CCA` to just implement `_transform_X` and `_transform_Y`
- Added `RDA` transformer (redundancy analysis) as another estimator that inherits from `ConstrainedOrdination`
- Added `GNNRDARegressor` as another GNN technique.  This needs to be further refactored in order to reduce duplication with `GNNRegressor`.

#### Issues
- While the design of `CCA` and `RDA` is reasonable (and two other ordination methods which follow this pattern can further be implemented), there is way too much duplication in the associated transformers (`CCATransformer` and `RDATransformer`) as well as NN estimators (`GNNRegressor` and `GNNRDARegressor`).  This particular combination (RDA + kNN) is not fully supported in `yaImpute` and, to my knowledge, hasn't been widely used as a mapping technique.  However, RDA can be used as a technique to compare two different multivariate datasets and, as such, can be used as a valid estimator.  We may want to think about providing the method (i.e. "cca" vs. "rda" vs. others yet to be implemented) as a hyperparameter to a `ConstrainedTransformer`.
- I think there may be some duplication with the new helper function called `is_2d_numeric_numpy_array` and constraints of the transformers themselves.   We may not need to be as stringent with all checks as the transformer would likely cover most cases.  
- I collapsed non-referenced properties in what was `OrdConstrained` (now `ConstrainedOrdination`) into local variables and only retained the properties that are used by the enclosing transformer.  This might limit the utility of these classes for other purposes, but we can always expose these local variables if needed.
- I've added a `__call__` method to `ConstrainedOrdination` as a means for setting the needed instance properties.  The return value of `__call__` is the instance itself which gets passed back to the enclosing transformer (e.g. `CCATransformer`).  Not sure if this is a good pattern or not.


grovduck:
@aazuspan, this is a bit of a mess right now, but wanted to get your eyes on it before I went too much further with it.  As I mention above, we can implement `RDA` and a couple of other ordination techniques fairly easily from the new `ConstrainedOrdination` superclass, but I think we'll want to think about how that cascades up to the transformer and then to the estimator.  Please let me know if my explanation isn't clear enough.  I've provided very scant documentation in the code as well, so I can enhance that if they are confusing.

Mainly, I wanted to get this committed and pushed so that I can look at your work on #59.

(Looks like I'm failing a documentation check, so I'll probably need a bit of hand-holding to know what I need to do to make that check pass.)

src/sknnr/transformers/_cca.py
```diff
@@ -13,228 +22,85 @@ def zero_sum_columns(arr: NDArray) -> NDArray:
...
-    @property
-    def site_wa_scores(self):
-        return self.WA
+class CCA(ConstrainedOrdination):
+    def _transform_Y(self):
+        normalized = self.Y / self.Y.sum()
+        self.rw = normalized.sum(axis=1)
+        self.cw = normalized.sum(axis=0)
+        rc = np.outer(self.rw, self.cw)
+        self.Y = (normalized - rc) / np.sqrt(rc)
```
aazuspan, line 91: I'm a little surprised that `_transform_X` returns the transformed X while `_transform_Y` sets `Y` attribute. Is that because `_transform_Y` mutates the class and should only be called once during initialization, while `_transform_X` can get called repeatedly? I hate to suggest renaming without offering any options, but maybe renaming one of them would be an easy way to make it clear that they work differently?

src/sknnr/transformers/_cca.py
```diff
@@ -13,228 +22,85 @@ def zero_sum_columns(arr: NDArray) -> NDArray:
...
+        U, S, _ = np.linalg.svd(Y_fit, full_matrices=False)
+        self.rank = min(rank, np.sum([S > self.ZERO]))
+        self.eigenvalues = np.square(S)[: self.rank]
+        self.axis_weights = np.diag(np.sqrt(self.eigenvalues / self.eigenvalues.sum()))
+        self.coefficients = np.linalg.lstsq(R, Q.T @ U[:, : self.rank], rcond=None)[0]
+        return self
 
     @property
     def max_components(self):
         return self.rank
```
aazuspan, line 76: What do you think about replacing `self.rank` with `self.max_components` to avoid needing this property as an "alias"? It would make `__call__` a little more wordy.

aazuspan (commented):
I'll take another pass tomorrow and try to look a little closer at some of the details, but from what I've seen this looks great! Huge improvement! As you mentioned, there's room for reducing duplication, but that hopefully shouldn't be too hard once we figure out the best approach (inheritance vs. parameters or something else). 

> I've added a __call__ method to ConstrainedOrdination as a means for setting the needed instance properties. The return value of __call__ is the instance itself which gets passed back to the enclosing transformer (e.g. CCATransformer). Not sure if this is a good pattern or not.

This is one of the areas I need to take a closer look at, but was there an advantage to this approach over setting everything through `__init__`? If I understand right, you would have to re-initialize the ordination to get a different result out of `__call__`, but I'm not clear if there might be a case where you would want to initialize without calling or call multiple times.

> I think there may be some duplication with the new helper function called is_2d_numeric_numpy_array and constraints of the transformers themselves. We may not need to be as stringent with all checks as the transformer would likely cover most cases.

Yeah, I noted something similar in one of my comments. I agree there's some redundancy between the validation checks, although I'm not sure whether it makes more sense to do the checks up front in the transformer or keep that complexity at a lower level in the ordination...

> This particular combination (RDA + kNN) is not fully supported in yaImpute and, to my knowledge, hasn't been widely used as a mapping technique.

Cool, it's exciting that we'll be able to offer additional functionality beyond `yaImpute`! Out of curiosity, how did you generate the test data?

> (Looks like I'm failing a documentation check, so I'll probably need a bit of hand-holding to know what I need to do to make that check pass.)

You can ignore that! I had to set up RTD to build docs for all PRs to make sure it was building correctly for #59, but because this branch doesn't have any config for the docs, RTD rightfully complains. I can disable that setting if it gets too annoying (it will trigger every new commit), but otherwise we can just ignore and merge with the failing check until the docs get merged in.

src/sknnr/_gnn_rda.py
```diff
@@ -0,0 +1,48 @@
+from sklearn.base import TransformerMixin
+
+from ._base import TransformedKNeighborsRegressor, YFitMixin
+from .transformers import RDATransformer
+
+
+class GNNRDARegressor(YFitMixin, TransformedKNeighborsRegressor):
```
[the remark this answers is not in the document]
aazuspan: Reading through your post more carefully, I think this is the same idea you were proposing for combining `CCATransformer` and `RDATransformer` into `ConstrainedTransformer`, right?

>  We may want to think about providing the method (i.e. "cca" vs. "rda" vs. others yet to be implemented) as a hyperparameter to a ConstrainedTransformer.

In both cases, having one parameterized class vs. two classes both seem like valid approaches to me, and I think we could end up with similarly clean code either way if we used inheritance to reduce duplication with the latter option. I'm happy to leave the decision up to if you have any preference either way!

src/sknnr/transformers/_cca.py
```diff
@@ -13,228 +22,85 @@ def zero_sum_columns(arr: NDArray) -> NDArray:
...
-        return np.square(self.sol[1])[: self.rank]
-
-    @property
-    def _V(self):
-        return self.sol[2].T[:, : self.rank]
-
-    @property
-    def U(self):
-        return self._U / np.sqrt(self.rw)[:, None]
+    def __init__(self, X, Y):
```
aazuspan, line 28: Apologies since I know we discussed this before but I couldn't find it: is there a rule/standard for when to use uppercase `Y` instead of `y`?

grovduck:
@aazuspan, thanks for the great review.  Very helpful comments (even on a pretty messy draft) and I think I've addressed most of your comments although there are a few still left to resolve.  I'm also thinking about just having this PR address the `CCA`/`RDA` side of things and then I'll start another PR on `CCorA`.  Otherwise, I think this will be too unwieldy.

A couple of comments below, but I mostly have inline responses to your review.

> > I've added a **call** method to ConstrainedOrdination as a means for setting the needed instance properties. The return value of **call** is the instance itself which gets passed back to the enclosing transformer (e.g. CCATransformer). Not sure if this is a good pattern or not.
> 
> This is one of the areas I need to take a closer look at, but was there an advantage to this approach over setting everything through `__init__`? If I understand right, you would have to re-initialize the ordination to get a different result out of `__call__`, but I'm not clear if there might be a case where you would want to initialize without calling or call multiple times.

No, you're right, there is no reason to have a `__call__` method in there.  I've removed it with this round and put it back into `__init__`.  It felt a bit "dirty" to do all the actual work in `__init__`, but I think that's preferable to an unfamiliar API.
 
> You can ignore that! I had to set up RTD to build docs for all PRs to make sure it was building correctly for #59, but because this branch doesn't have any config for the docs, RTD rightfully complains. I can disable that setting if it gets too annoying (it will trigger every new commit), but otherwise we can just ignore and merge with the failing check until the docs get merged in.

Sounds good.  My guess is that we'll get #59 in place first and then I'll merge those changes into this branch.

src/sknnr/_gnn_rda.py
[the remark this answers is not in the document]
grovduck: > Reading through your post more carefully, I think this is the same idea you were considering for combining `CCATransformer` and `RDATransformer` into `ConstrainedTransformer`, right?

Yes, this was my initial thought and it's the way I currently have it implemented.  `CCATransformer` and `RDATransformer` have now been replaced with just `ConstrainedTransformer` which now takes a `method` keyword argument (currently "cca" or "rda" and defaulting to "cca").  There's a bit of logic to test for and initialize the correct `ConstrainedOrdination` subclass, but that's really the only change between what was `CCATransformer` and `ConstrainedTransformer`.

Likewise, `GNNRegressor` now has the same `method` keyword argument and internally its transformer is just a `ConstrainedTransformer` instance (rather than a `CCATransformer` instance).

Does this seem like a valid approach?

Probably, the most substantive change I made was to place `ConstrainedOrdination`, `CCA`, and `RDA` all into their own separate files.  I'm not really sure this is necessary and am fine if they all live in the same file (probably `transformers/_constrained_ordination.py`).

src/sknnr/_gnn_rda.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def _more_tags(self):
+        unsupported_1d = "RDA requires 2D y arrays."
+
+        return {
+            "allow_nan": False,
+            "requires_fit": True,
+            "requires_y": True,
+            "multioutput": True,
+            "_xfail_checks": {
```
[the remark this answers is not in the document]
grovduck: I wasn't *exactly* sure what you meant here, so I haven't yet done anything here.  Are you suggesting that the dictionary is something like:
```
unsupported_1d = "Constrained ordination requires 2D y arrays."
UNSUPPORTED_1D_CHECKS = dict(
    check_estimator_dtypes=unsupported_1d,
    check_dtype_object=unsupported_1d,
    ...
)
```
and then in `GNNRegressor` and (now) `ConstrainedOrdination`, we do:
```
def _more_tags(self):
    return {
        "allow_nan": False,
        ...
        "_xfail_checks": UNSUPPORTED_1D_CHECKS
    }
```
I'm a bit confused because the check names aren't exactly the same between the two classes.

src/sknnr/transformers/_cca.py
```diff
@@ -13,228 +22,85 @@ def zero_sum_columns(arr: NDArray) -> NDArray:
...
-    def colsum(self):
-        return self.transform.cw
-
-
-class ConstrainedOrdination:
-    Y_TRANSFORM = {
-        "CCA": InitCA,
-    }
+class ConstrainedOrdination(ABC):
     ZERO: float = math.sqrt(2.220446e-16)
```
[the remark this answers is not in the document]
grovduck: Good catch.  Yes, definitely a relic from yaImpute.  I've taken it out and tests are passing so I'm going to assume we're OK.

src/sknnr/transformers/_cca.py
```diff
@@ -13,228 +22,85 @@ def zero_sum_columns(arr: NDArray) -> NDArray:
...
-    @property
-    def _V(self):
-        return self.sol[2].T[:, : self.rank]
-
-    @property
-    def U(self):
-        return self._U / np.sqrt(self.rw)[:, None]
+    def __init__(self, X, Y):
+        self.X = np.array(X)
+        self.Y = np.array(Y)
```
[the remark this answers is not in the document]
grovduck: Perfect.  Replaced here and also in `CCorA`.

src/sknnr/transformers/_cca.py
```diff
@@ -1,10 +1,19 @@
...
+from abc import ABC, abstractmethod
+from typing import Any
 
 import numpy as np
 from numpy.typing import NDArray
 
 
+def is_2d_numeric_numpy_array(arr: Any) -> bool:
+    return (
+        isinstance(arr, np.ndarray)
```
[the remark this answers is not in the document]
grovduck: Yep, you're absolutely right and I don't think we necessarily need it elsewhere.  I've renamed the function to `is_2d_numeric_array` and now `arr` is typed as `NDArray`.

src/sknnr/transformers/_cca.py
```diff
@@ -13,228 +22,85 @@ def zero_sum_columns(arr: NDArray) -> NDArray:
...
+        ...
+
+    @abstractmethod
+    def _transform_X(self):
+        ...
+
+    def _check_inputs(self):
+        if not is_2d_numeric_numpy_array(self.X):
+            msg = "X must be a 2D numeric numpy array"
+            raise ValueError(msg)
```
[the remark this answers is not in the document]
grovduck: Again, I *think* I understand what you meant here - basically set the `msg` in an if/elif ladder and then raise at the end?  I went ahead and got rid of the superfluous msg variable in each for now and just am raising the string literal.  Please definitely feel free to rewrite in a more concise way.  My brain wasn't seeing it today ...

src/sknnr/transformers/_cca.py
```diff
@@ -13,228 +22,85 @@ def zero_sum_columns(arr: NDArray) -> NDArray:
...
+    def _transform_X(self):
+        self.env_center = np.average(self.X, axis=0, weights=self.rw)
+        X_scale = self.X - self.env_center
+        return X_scale * np.sqrt(self.rw)[:, np.newaxis]
 
 
-class CCA(ConstrainedOrdination):
-    method = "CCA"
-    inertia = "scaled Chi-square"
+class RDA(ConstrainedOrdination):
```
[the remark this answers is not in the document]
grovduck: Definitely.  Although as I mentioned elsewhere, I have separate modules for `ConstrainedOrdination`, `CCA`, and `RDA` and I'm second-guessing that decision.  I think we could have `CCA` and `RDA` in the same file just trailing `ConstrainedOrdination` instead.  Let me know what you think.

src/sknnr/transformers/_cca_transformer.py
```diff
@@ -72,3 +72,71 @@ def _more_tags(self):
                 "check_requires_y_none": unsupported_1d,
             },
         }
+
+
+class RDATransformer(ComponentReducerMixin, TransformerMixin, BaseEstimator):
```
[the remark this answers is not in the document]
grovduck: Correct!  This file has now gone away, replaced with the generic `ConstrainedTransformer` class in the file `transformers/_constrained_transformer.py`.

src/sknnr/transformers/_cca_transformer.py
```diff
@@ -72,3 +72,71 @@ def _more_tags(self):
...
+            X,
+            reset=True,
+            dtype=FLOAT_DTYPES,
+            force_all_finite=True,
+            ensure_min_features=2,
+            ensure_min_samples=1,
+        )
+        y = np.asarray(y)
+        if len(y.shape) < 2:
+            raise ValueError("`y` must be a 2D array.")
```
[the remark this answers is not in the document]
grovduck: Good catch!  Removed and tested!

src/sknnr/transformers/_cca.py
grovduck: You're absolutely right on this.  For now, both `_transform_X` and `_transform_Y` resets the `self.X` and `self.Y` arrays within the functions.  I don't love this.  In addition to the side effects of setting these within these methods, we're also setting other instance attributes (e.g. `env_center`, `rw`, and `cw` in `CCA` and `env_center` in `RDA`).  I'd love your opinion on where it would be best to set these instance attributes.  Do you feel like it's OK to hide within these methods?

src/sknnr/transformers/_cca.py
grovduck: I hemmed and hawed on this one for a while.  I don't think the latter suggestion works because `CCorA` doesn't return `rank` as its `max_components` but rather the number of "significant" components.  But the former suggestion would definitely work.  I haven't done so and my logic might be a bit shaky, so see what you think.

I've kept the term `rank` in there to be as consistent as possible with the variable names in `vegan` and because `rank` would be the default name when running `np.linalg.lstsq` (even though we actually modify it later on for positive eigenvalues).  I felt like the `max_components` property was lightweight and inobtrusive enough to have it alias `rank` but not completely replace it.  Clean-code-wise, it definitely makes sense to do as you say, but having `rank` in there might be a valuable reminder for me.

src/sknnr/transformers/_cca.py
grovduck: Again, this is a difference in terminology between `sklearn`  and `vegan` conventions.  Because `ConstrainedOrdination` is more or less a port of vegan, I felt like keeping the matrix variables upper-cased was important to follow those conventions.   I'm not as strident on this one, though, and could be convinced to change `Y` to `y`.

grovduck:
Also, I'd love your review of `test_port.py` on this go around.  We now have two reasons to "uncollect" tests:

1. If an estimator does not support `n_components` and a value for `n_components` was specified
2. If an estimator does not support `method` and a value for `constrained_method` was specified [^1]

For 2 though it gets even a bit more tricky in that we want to exclude the `GNNRegressor` tests that *don't* pass a `constrained_method`.  So we really need to uncollect if the estimator supports `method` *and* a value for `constrained_method` was not specified.  

There is also the issue that only one function can be specified for the `pytest.mark.uncollect_if` marker, so now there is a new function called `estimator_does_not_support_parametrization` that "or"s the two other conditions.

You'll also see that I've renamed the test data files called "gnn_moscow_*" to "gnn_cca_moscow_*" so that the `method` variable finds the correct dataset when calling `load_moscow_stjoes_results`.  There's definitely some duplicated logic across the tests that I'm not particularly happy with, so if you have suggestions there, I'd love to hear it.

[^1]: You'll clearly note the inconsistency in naming here - does it make sense to change `method` to `constrained_method` in `GNNRegressor` and `ConstrainedTransformer`?

src/sknnr/transformers/_cca.py
aazuspan: I might suggest going lowercase for consistency if this was in the public API, but since it's just an implementation within the transformer, sticking with the `vegan` conventions is fine with me!

src/sknnr/_gnn_rda.py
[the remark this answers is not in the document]
aazuspan: Nice, that refactor cleaned up the duplication a lot! I think as long as the difference between `CCA` and `RDA` feels minor enough to you for it to be a parameter, that's probably the right call. The downside is having to validate the user string, but I think it's probably worth the reduced complexity versus using inheritance.

Regarding the file organization, I have a slight preference for combining them into a single file, but could be persuaded if you disagree.

src/sknnr/_gnn_rda.py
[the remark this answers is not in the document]
aazuspan: That's what I had in mind. I think the primary difference between the two sets of checks was that different tests get collected for transformers vs. estimators and based on their tags, although I can't say I have a great understanding of that. It's probably not worth the complexity if we have to keep two separate sets of checks to xfail. We could just import a `dict` of shared checks and manually add the rest, but I'm not sure that's an improvement either. I'm fine to leave as-is for now, although I keep hoping we'll find a way to get rid of those some day...

src/sknnr/transformers/_cca.py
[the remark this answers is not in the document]
aazuspan: That was my suggestion, but I think your solution is just as readable, and a little more concise.

src/sknnr/_gnn_rda.py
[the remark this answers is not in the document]
grovduck: > Regarding the file organization, I have a slight preference for combining them into a single file, but could be persuaded if you disagree.

Yes, I was leaning that way myself and will make this change.

> I think as long as the difference between CCA and RDA feels minor enough to you for it to be a parameter, that's probably the right call. The downside is having to validate the user string, but I think it's probably worth the reduced complexity versus using inheritance.

Actually, I'm now a bit curious how you would implement this via inheritance?  I'm not seeing that pattern.  If it's not too much trouble to sketch it out and you'd prefer that design, I'm happy to consider it.

aazuspan:
> I'm also thinking about just having this PR address the CCA/RDA side of things and then I'll start another PR on CCorA.

Sounds good!

> My guess is that we'll get https://github.com/lemma-osu/scikit-learn-knn-regression/pull/59 in place first and then I'll merge those changes into this branch.

Yeah, this is a good idea since we'll need to make some changes to the API Reference for the renamed `CCATransformer`.

> Also, I'd love your review of test_port.py on this go around.

Definitely - I'll take a close look at the testing side tomorrow!

src/sknnr/transformers/_cca.py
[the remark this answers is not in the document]
grovduck: The module `_constrained_ordination` now has `ConstrainedOrdination`, `CCA`, and `RDA` in it.

grovduck:
> Out of curiosity, how did you generate the test data?

I forgot to answer this question.  It's an ugly mess, but I created local copies of `yaImpute` functions (specifically `yai` and `newtargets`) and added a "gnn_rda" section that uses a call to vegan's `rda` instead of `cca`.  If/when I finish this off, I can do the same with the other two methods: `capscale` and `dbrda`, but need to be a bit more confident that I know generally how they differ from `cca`/`rda`.

src/sknnr/transformers/_cca.py
aazuspan: Cool, I think this is an improvement. The side effects of those methods, and the fact they have to be called in the right order does make me a little uncomfortable, too, though.

Maybe one minor improvement could be to have the `_transform` methods take `X` and `Y` parameters, rather than using `self.X` and `self.Y` (although they would still *set* those attributes). That would prevent the possibility of somehow calling them twice and double-transforming the `X` and `Y`.

The other thought is that maybe they could be renamed to at least make it clear that they will have side-effects, e.g. `_set_transformed_X(X: np.NDArray) -> None`. That doesn't help the fact that they also set other attributes or that they need to be called in the right order, so I guess another option would be to combine the two methods into one method with a generic name like `_initialize_ordination` or `_set_initialization_attrs` or something, which would take `X` and `Y` params.

aazuspan: Cool, I think this is an improvement. The side effects of those methods, and the fact they have to be called in the right order does make me a little uncomfortable, too, though.

Maybe one minor improvement could be to have the `_transform` methods take `X` and `Y` parameters, rather than using `self.X` and `self.Y` (although they would still *set* those attributes). That would prevent the possibility of somehow calling them twice and double-transforming the `X` and `Y`.

The other thought is that maybe they could be renamed to at least make it clear that they will have side-effects, e.g. `_set_transformed_X(X: np.NDArray) -> None`. That doesn't help the fact that they also set other attributes or that they need to be called in the right order, so I guess another option would be to combine the two methods into one method with a generic name like `_initialize_ordination` or `_set_initialization_attrs` or something, which would take `X` and `Y` params.

src/sknnr/_gnn_rda.py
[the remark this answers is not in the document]
grovduck: I agree that it's "unsightly" and it would be nice to squirrel it away somewhere that we didn't have it so present, but I think I'm on board with keeping it as-is until we can somehow magically pass all those checks.

src/sknnr/transformers/_constrained_transformer.py
```diff
@@ -16,11 +25,14 @@ def fit(self, X, y):
...
-        if len(y.shape) < 2:
-            raise ValueError("`y` must be a 2D array.")
 
-        self.ordination_ = CCA(X, y)
+        if self.method == "cca":
+            self.ordination_ = CCA(X, y)
+        elif self.method == "rda":
+            self.ordination_ = RDA(X, y)
+        else:
+            raise ValueError("`method` must be one of 'cca' or 'rda'.")
```
aazuspan, line 34: I can see why you put this check here since this is where we need to instantiate the ordination, but if possible it would be nice to make this check in `__init__` so that it fails as soon as possible.

Just to throw out a few untested options, what do you think about something like below?

```python
CONSTRAINED_METHODS = {
    "cca": CCA,
    "rda": RDA,
}

class ConstrainedTransformer(ComponentReducerMixin, TransformerMixin, BaseEstimator):
    def __init__(
        self, method: Literal["cca", "rda"] = "cca", n_components: int | None = None
    ):
        self.method = method
        if self.method not in CONSTRAINED_METHODS:
            raise ValueError(f"`method` must be one of {CONSTRAINED_METHODS.keys()}, not {method}.")
        
        super().__init__(n_components=n_components)

    def fit(self, X, y):
        X = self._validate_data(
            X,
            reset=True,
            dtype=FLOAT_DTYPES,
            force_all_finite=True,
            ensure_min_features=2,
            ensure_min_samples=1,
        )
        self.ordination_ = CONSTRAINED_METHODS[self.method](X, y)
        self.set_n_components()
        self.env_center_ = self.ordination_.env_center
        self.projector_ = self.ordination_.projector(n_components=self.n_components_)
        return self
```

Or we could set the ordination class in `__init__` instead of the name. This is a little cleaner, but maybe it's too weird to take a parameter and set a different attribute with the same name?

```python
class ConstrainedTransformer(ComponentReducerMixin, TransformerMixin, BaseEstimator):
    def __init__(
        self, method: Literal["cca", "rda"] = "cca", n_components: int | None = None
    ):
        self.method = CONSTRAINED_METHODS.get(method)
        if self.method is None:
            raise ValueError(f"`method` must be one of {CONSTRAINED_METHODS.keys()}, not {method}.")
        
        super().__init__(n_components=n_components)

    def fit(self, X, y):
        X = self._validate_data(
            X,
            reset=True,
            dtype=FLOAT_DTYPES,
            force_all_finite=True,
            ensure_min_features=2,
            ensure_min_samples=1,
        )
        self.ordination_ = self.method(X, y)
        self.set_n_components()
        self.env_center_ = self.ordination_.env_center
        self.projector_ = self.ordination_.projector(n_components=self.n_components_)
        return self
```

grovduck: I like either of these patterns ... much better to use a dictionary here.  A couple of questions:

* For constants (dicts or otherwise), I'm never clear when to make them module- or class-level.  Is there an advantage to having `CONSTRAINED_METHODS` as a module-level constant?
* In your second solution, is there a reason why the user argument (`method`) and the attrribute need to be named the same?  Would it make sense to call the attribute something like `self.method_cls`?

I yield to your judgement on what you think is best.  Both solutions are better than I what I currently have and neither is very confusing to me.

src/sknnr/transformers/_cca.py
grovduck: Great points, here.  I'll noodle on this, but definitely passing both `X` and `Y` together make much more sense.  So, I assume you're thinking about something like:
```
class ConstrainedOrdination(ABC):
    def __init__(self, X, Y):
        X = np.asarray(X)
        Y = np.asarray(Y)
        self._check_inputs(X, Y)
        self._set_initialization_attrs(X, Y) # this would be @abstractmethod
```

src/sknnr/transformers/_constrained_transformer.py
aazuspan: > Is there an advantage to having CONSTRAINED_METHODS as a module-level constant?

No, I think you're right that this should be class-level!

> In your second solution, is there a reason why the user argument (method) and the attrribute need to be named the same? Would it make sense to call the attribute something like self.method_cls?

I had a vague memory that `sklearn` prevents setting attributes that don't match the `__init__` args, but while [reading up on that](https://scikit-learn.org/stable/developers/develop.html#instantiation), I found that they recommend against having any *logic* in `__init__`:

> There should be no logic, not even input validation, and the parameters should not be changed [in `__init__`]. The corresponding logic should be put where the parameters are used, typically in fit

Looks like you had it right from the beginning!

src/sknnr/transformers/_cca.py
aazuspan: Yeah, that was what I had in mind.

src/sknnr/_gnn_rda.py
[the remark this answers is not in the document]
aazuspan: I think this could be done with inheritance by keeping a `ConstrainedTransformer` class like you currently have (minus the `method` parameter) and having `CCATransformer` and `RDATransformer` subclasses that only override an abstract `fit` with their respective ordinations. 

But I don't have any preference for that over the solution you went with, especially if we're going to end up with more constrained methods.

aazuspan:
> Also, I'd love your review of test_port.py on this go around.

Thanks for the thorough explanation of the updated uncollecting system! I think your solution looks great, considering all the complexity of different cases we have to cover.

It's possible there's some way we could clean up the data loading and parameterization system now that we have a better idea of what we need, but I don't have any specific ideas, and I'm thinking it's probably not worth the time or effort to do a big redesign when it's ultimately going to be replaced by #42.

> There is also the issue that only one function can be specified for the pytest.mark.uncollect_if marker, so now there is a new function called estimator_does_not_support_parametrization that "or"s the two other conditions.

This seems like a good workaround. I suspect we could probably modify `pytest_collection_modifyitems` to work with multiple `uncollect_if` markers by iterating over all markers instead of using `get_closest_marker`, but that seems like a lot of trouble for very little benefit, especially given that our port tests are going to eventually be replaced by #42.

> You'll clearly note the inconsistency in naming here - does it make sense to change method to constrained_method in GNNRegressor and ConstrainedTransformer?

I'm fine with either! I don't think that we should change it for the sake of our tests, but it is a more descriptive name, so I don't think it would hurt to switch.

To help with the confusion between the `method`/`constrained_method` in the tests, I think we *could* modify the test code to refer to the estimator name as `name` or `id` or something else instead of `method`, but I don't have a strong feeling on that.

src/sknnr/_gnn_rda.py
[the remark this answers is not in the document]
grovduck: Sorry to drag out this discussion, but wouldn't we essentially be "kicking the can" of changing which *transformer* to use (e.g. `CCATransformer`, `RDATransformer`) in `GNNRegressor` based on `method` versus changing which *ordination* to use (e.g. `CCA`, `RDA`) in `ConstrainedTransformer` based on `method`?  I feel like I'm not seeing something ...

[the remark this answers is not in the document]
aazuspan: Yeah, I don't think there's any benefit to the implementation from keeping `CCATransformer` and `RDATransformer` separate. The only meaningful difference would be the public API, but I don't think that's necessarily a benefit either.

src/sknnr/transformers/_cca.py
grovduck: I've made this change.  The one change I'm not thrilled with is that I'm passing `X` and `Y` back from `self._check_inputs` because `Y` can be modified based on if there are zero-sum columns.  Even though there is no chance that `X` gets modified, it felt weird just returning `Y` and I didn't want to set `self.Y` in `_check_inputs`.  Let me know what you think of this.  The other thing I just realized is that we can make `_check_inputs` a `staticmethod`.

src/sknnr/transformers/_constrained_transformer.py
grovduck: OK, I think I have this right, although it feels weird that we can initialize `ConstrainedTransformer` with a bogus `constrained_method` and it won't error until fitting.  Here's the relevant bits:
```
class ConstrainedTransformer(ComponentReducerMixin, TransformerMixin, BaseEstimator):
    CONSTRAINED_METHODS = {
        "cca": CCA,
        "rda": RDA,
    }

    def __init__(
        self,
        constrained_method: Literal["cca", "rda"] = "cca",
        n_components: int | None = None,
    ):
        self.constrained_method = constrained_method
        super().__init__(n_components=n_components)

    def fit(self, X, y):
        ...
        method_cls = self.CONSTRAINED_METHODS.get(self.constrained_method)
        if method_cls is None:
            raise ValueError(
                f"`method` must be one of {self.CONSTRAINED_METHODS.keys()}, not"
                f" {self.constrained_method}."
            )
        self.ordination_ = method_cls(X, y)
```
Again, if this isn't quite right, I'm happy to change!

grovduck:
> It's possible there's some way we could clean up the data loading and parameterization system now that we have a better idea of what we need, but I don't have any specific ideas, and I'm thinking it's probably not worth the time or effort to do a big redesign when it's ultimately going to be replaced by #42.

I might be confused, but I think we'll still have to have the uncollect system in place to run the right combinations, but we won't have to build the correct estimator name in order to fetch the yaImpute-based files.  Is that right?  Or are you thinking that we won't test all combinations once we have regression testing in place?



> I'm fine with either! I don't think that we should change it for the sake of our tests, but it is a more descriptive name, so I don't think it would hurt to switch.
> 
> To help with the confusion between the `method`/`constrained_method` in the tests, I think we _could_ modify the test code to refer to the estimator name as `name` or `id` or something else instead of `method`, but I don't have a strong feeling on that.

I decided to change `method` to `constrained_method` in `GNNRegressor` and `ConstrainedOrdination` even though it felt a bit redundant to have `ConstrainedOrdination.constrained_method`.  I kept the tests more or less the same.

Some other very small changes:
* Changed `is_2d_numeric_array` to have required keyword arguments
* Combined `zero_sum_rows` and `zero_sum_columns` to `zero_sum_vectors` with an `axis` parameter.  It too has required keyword arguments.

That feels safer to me, but it might be too strigent?

src/sknnr/transformers/_cca.py
aazuspan: Yeah, I can see how it's a little odd to return `X` unchanged, but I agree with your logic that the alternatives would be weirder. This looks good to me!

src/sknnr/transformers/_constrained_transformer.py
aazuspan: > it feels weird that we can initialize ConstrainedTransformer with a bogus constrained_method and it won't error until fitting

I agree, but you can do the same thing with some `sklearn` estimators like `criterion` for `RandomForestRegressor`. I guess the downside of building off of `sklearn` is that we inherit some of their questionable choices.

I'm going to suggest one minor change in an inline comment, but otherwise this looks great!

src/sknnr/transformers/_constrained_transformer.py
```diff
@@ -16,11 +32,15 @@ def fit(self, X, y):
...
         )
-        y = np.asarray(y)
-        if len(y.shape) < 2:
-            raise ValueError("`y` must be a 2D array.")
 
-        self.ordination_ = CCA(X, y)
+        method_cls = self.CONSTRAINED_METHODS.get(self.constrained_method)
+        if method_cls is None:
+            raise ValueError(
+                f"`method` must be one of {self.CONSTRAINED_METHODS.keys()}, not"
```
aazuspan, line 39:
```suggestion
                f"`method` must be one of {list(self.CONSTRAINED_METHODS.keys())}, not"
```

My bad for suggesting this originally! As is, it will print `method must be one of dict_keys(['cca', 'rda'])`.

aazuspan:
> I might be confused, but I think we'll still have to have the uncollect system in place to run the right combinations, but we won't have to build the correct estimator name in order to fetch the yaImpute-based files. Is that right?

I hadn't thought that far ahead, apparently! Of course you're right that we'll still need to parameterize and uncollect as needed.

> Changed is_2d_numeric_array to have required keyword arguments

My personal preference with required kwargs is to use them when it isn't clear what the args should be or when the order is arbitrary, so if it were just me I would probably keep `arr` positional but make `axis` a required kwarg. But that's just me! It certainly doesn't hurt to require them.

>Combined zero_sum_rows and zero_sum_columns to zero_sum_vectors with an axis parameter

Good call!

grovduck:
> My personal preference with required kwargs is to use them when it isn't clear what the args should be or when the order is arbitrary, so if it were just me I would probably keep `arr` positional but make `axis` a required kwarg. But that's just me! It certainly doesn't hurt to require them.

I agree with your advice on this.  I've reverted these functions such that `arr` is positional and `axis` is a required kwarg.

src/sknnr/transformers/_constrained_transformer.py
grovduck: Thanks for the fix.  I've changed this.

grovduck:
Aaaand it looks like we'll need a docs change already :wink:.  I'll address that once I decide about adding the other two methods for `ConstrainedOrdination`.

https://readthedocs.org/projects/sknnr/builds/22259462/

PR #62, opened 2023-10-17
Fix README doc links

aazuspan:
Quick fix for the links to the docs in README (for future reference, we need to link to `https://sknnr.readthedocs.io/en/latest/***`).


grovduck:
I had to remind myself what the issue was, but it looks like you just discovered that `en/latest` is part of the URI.  LGTM!

PR #63, opened 2023-10-25
Add `3.12` to CI

aazuspan:
The biggest change is adding `3.12` to our CI now that it's officially out (we're passing!), but I also bumped a few actions to the latest versions and added a docs badge to the readme while I was at it. @grovduck I hope you don't mind the combined PR!

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,16 +8,16 @@ jobs:
   test:
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
 
     runs-on: ubuntu-latest
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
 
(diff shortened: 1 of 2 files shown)


grovduck (approved):
All looks good to me.  Out of curiosity, when do you typically check for new Github action versions and is this something that needs to be manually updated (rather than pointing to a `@latest` release)?

aazuspan:
Thanks for giving it a look!

My system for updating actions is to completely forget about them until I'm making an unrelated change to the CI, then bump them to the latest major release and cross my fingers :stuck_out_tongue_winking_eye:. I've never had an issue or noticed any change when I update, so I suspect they're mostly pretty stable.

You *can* get the latest release by pointing to a specific branch, e.g. `actions/checkout@main`, but the [Github docs](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runsstepsuses) suggest pinning to a major release to avoid unexpected breaking changes when there's a new major release.

grovduck:
> My system for updating actions is to completely forget about them until I'm making an unrelated change to the CI, then bump them to the latest major release and cross my fingers 😜

Ahh, so remarkably close to my system!

PR #65, opened 2024-03-20
Rename package to sknnr and prep for alpha release

### LINKED ISSUE #61: Change installed name to `sknnr`
Our previous plan was to use `scikit-learn-knn-regression` as the installed name of the package and `sknnr` as the imported name, but per the [discussion](https://github.com/lemma-osu/scikit-learn-knn-regression/pull/59#discussion_r1344430580) in #59, we're now tentatively planning to use `sknnr` across the board.

### Pros for switching to `sknnr`:
- Shorter, catchier, more memorable
- No confusion about what name to install and/or import

### Cons for switching to `sknnr`: 
- Less descriptive
- No explicit reference to `scikit-learn`

Making the switch will be pretty straightforward, and should just require updating the `pyproject.toml`, the repository name on Github, and links to the repository throughout the docs (although Github will continue to redirect).

If we do make a final decision to go with `sknnr` on PyPI, I suggest we make an alpha release just to claim the package name (pre-releases can only be installed with `pip install sknnr --pre`, so it should be clear that it is not an official, stable release if anyone comes across it).

aazuspan:
Closes #61 by updating all references from `scikit-learn-knn-regression` to `sknnr`. Also preps for the `0.1.0-alpha` release by updating install instructions and the package version.


grovduck (approved):
Looks great, @aazuspan.  Out of curiosity, did you use `hatch version 0.1.0-alpha` to change the version or edit `__about__.py` directly?

aazuspan:
> Out of curiosity, did you use hatch version 0.1.0-alpha to change the version or edit __about__.py directly?

Good question. Based on pypa/hatch#697 it looked like setting a non-standard version wasn't going to be straightforward with `hatch`, so I just edited it manually.

PR #67, opened 2024-03-22
Update pre-commit versions and contributing doc

grovduck:
Per changes made in [`sknnr-spatial`](https://github.com/lemma-osu/sknnr-spatial/commit/1da7ceaf4af1384228a704859d126cc08855752b), update the pre-commit versions to latest versions. 
- `ruff: v0.3.2`
- `mypy: v1.9.0`

In addition, remove the use of `black` as a pre-commit hook and instead rely on `ruff-format`.  Also, update `contributing.md` to show command use with `hatch` and to align it with the contributing guide for `sknnr-spatial`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,8 +35,8 @@ dependencies = ["pre-commit"]
 dependencies = ["pytest", "pytest-cov", "pandas"]
 
 [tool.hatch.envs.test.scripts]
-all = "pytest {args}"
-coverage = "pytest --cov=src/sknnr {args}"
+all = "pytest . {args} --doctest-modules"
+coverage = "pytest . --cov=src/sknnr {args} --doctest-modules"
 
 [tool.hatch.envs.docs]
 dependencies = ["mkdocs", "mkdocs-material", "mkdocstrings[python]"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,13 +11,9 @@ license = ""
 requires-python = ">=3.8"
 authors = [
     { name = "Matt Gregory", email = "matt.gregory@oregonstate.edu" },
-    { name = "Aaron Zuspan", email = "aaron.zuspan@oregonstate.edu" }
+    { name = "Aaron Zuspan", email = "aaron.zuspan@oregonstate.edu" },
 ]
-dependencies = [
-    "numpy",
-    "scikit-learn>=1.2",
-    "scipy",
-]
+dependencies = ["numpy", "scikit-learn>=1.2", "scipy"]
 
 [project.urls]
 Homepage = "https://github.com/lemma-osu/sknnr"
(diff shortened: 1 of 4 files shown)


grovduck:
@aazuspan, a few notes on this implications of this commit:
- My editor did a lot of reformatting of the `pyproject.toml` file which I attribute to the swap of `black` for `ruff-format`?  The reformatting looks correct to me based on `black` formatting rules, but hopefully it doesn't cause you grief.
- There were warnings from `ruff` that configuration associated with `select` and `isort` now need to be under `ruff.lint` headings rather than `ruff` itself.
-  `mypy` brought out what it considered to be a type inconsistency on `sknnr.datasets._base._open_text`.  I've changed this to return `IO[str]` rather than `TextIO`.  Some discussion of this [here](https://github.com/python/typeshed/issues/283).

aazuspan (approved):
Looks great @grovduck, thanks for updating the pre-commits and getting everything back in order!

I think I had omitted the `hatch run` from the pre-commit instructions with the thought that it's not needed if you've activated the shell, but now it looks like the instructions will work whether or not the shell is activated, so definitely an improvement!

> My editor did a lot of reformatting of the pyproject.toml file which I attribute to the swap of black for ruff-format?

I'm not sure that `ruff-format` will work on `toml` files - I tried manually running `ruff-format` on the file and it just skipped it. But whatever made the changes, they look better to me!

> I've changed this to return IO[str] rather than TextIO. Some discussion of this https://github.com/python/typeshed/issues/283.

Good to know, thanks for the reference!

pyproject.toml
```diff
@@ -30,32 +26,20 @@ path = "src/sknnr/__about__.py"
...
 
 [tool.hatch.envs.test]
-dependencies = [
-    "pytest",
-    "pytest-cov",
-    "pandas"
-]
+dependencies = ["pytest", "pytest-cov", "pandas"]
 
 [tool.hatch.envs.test.scripts]
```
aazuspan, line 37: You could add `--doctest-modules` to the test scripts while you're making updates, but I don't think there are any doctests to run, so it would just be pre-emptive. Your call!

grovduck:
> I'm not sure that `ruff-format` will work on `toml` files - I tried manually running `ruff-format` on the file and it just skipped it

Ah, thanks for that clarification.  Then I'm guessing it must be a VSCode TOML extension I'm using (`Even Better TOML`).  As long as you don't mind, I'll keep the changes.

aazuspan:
That makes sense - yeah, I'd say keep the changes! We could add a TOML formatter to the pre-commit, but it changes so infrequently I'm not sure it's the effort.

pyproject.toml
grovduck: Good call!  I've added that to `pyproject.toml`

PR #69, opened 2024-07-10
Expose functions from `datasets` module in public API

### LINKED ISSUE #68: Expose dataset loading functions as public API
There are functions in `sknnr.datasets._base` that manage the loading of CSV files into `Dataset` objects (similar to sklearn's `Bunch` objects).  At present, however, these utility functions (e.g. `_load_dataset_from_csv_filenames` and `_load_csv_data`) are meant for internal use and are currently used in `load_moscow_stjoes` and `load_swo_ecoplot`.  As @aazuspan points out in lemma-osu/synthetic-knn#2, these functions (and callees) rely on "specific restrictions like requiring an integer index in the first column, and floats in the remaining columns" (see [here](https://github.com/lemma-osu/sknnr/blob/main/src/sknnr/datasets/_base.py#L84-L85)).

Nonetheless, even with these restrictions, this functionality is likely useful across other packages that require `sknnr` or for users that may have similarly formatted CSV files.  In order to use these functions in a public API, we would need to generalize `_load_csv_data` so that it possibly accepts `DATA_MODULE` as a keyword argument (as in  [`sklearn.datasets.load_csv_data`](https://github.com/scikit-learn/scikit-learn/blob/c44457af96a924683b6a65bd3fd6c285043bdec6/sklearn/datasets/_base.py#L318)).  This would work well for other packages that want to use this functionality, although perhaps users might have to create a module structure with their data to use it as well.

To consider:

- Should these data loading functions continue to live in `sknnr.datasets` or in a more generic "utilities" module?
- Do we need to support loading not only from modules (using `importlib.resources`) or from generic directories as well?

@aazuspan, further thoughts on this?

grovduck:
Closes #68 by adding `load_csv_data` and `load_dataset_from_csv_filenames` to the public API.  `data_module` is now passed as a keyword argument to both functions and defaults to `sknnr.datasets.data`.

- `data_module` in both functions can be passed as either a string (e.g. "sknnr.datasets.data") or as a module object
- Rudimentary tests on `load_csv_data`

### DIFF
--- a/src/sknnr/datasets/_base.py
+++ b/src/sknnr/datasets/_base.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import csv
 import sys
+import types
 from dataclasses import dataclass
 from importlib import resources
 from typing import IO, TYPE_CHECKING
@@ -54,7 +55,7 @@ def _dataset_as_frame(dataset: Dataset) -> Dataset:
     )
 
 
-def _open_text(module_name: str, file_name: str) -> IO[str]:
+def _open_text(module_name: str | types.ModuleType, file_name: str) -> IO[str]:
     """Open a file as text.
 
     This is a compatibility port for importlib.resources.open_text, which is deprecated
@@ -67,9 +68,7 @@ def _open_text(module_name: str, file_name: str) -> IO[str]:
 
 
 def load_csv_data(
-    file_name: str,
-    *,
-    data_module: str = DATA_MODULE,
+    file_name: str, *, data_module: str | types.ModuleType = DATA_MODULE
 ) -> tuple[NDArray[np.int64], NDArray[np.float64], NDArray[np.str_]]:
     """Load data from a CSV file from the specified data_module.
 
@@ -112,7 +111,7 @@ def load_dataset_from_csv_filenames(
     target_filename: str,
     return_X_y: bool = False,
     as_frame: bool = False,
-    data_module: str = DATA_MODULE,
+    data_module: str | types.ModuleType = DATA_MODULE,
 ) -> tuple[NDArray[np.float64], NDArray[np.float64]] | Dataset:
     """Load separate data and target CSV files into a dataset or paired NumPy arrays.
 
--- a/src/sknnr/datasets/_base.py
+++ b/src/sknnr/datasets/_base.py
@@ -68,15 +68,15 @@ def _open_text(module_name: str | types.ModuleType, file_name: str) -> IO[str]:
 
 
 def load_csv_data(
-    file_name: str, *, data_module: str | types.ModuleType = DATA_MODULE
+    file_name: str, *, module_name: str | types.ModuleType = DATA_MODULE
 ) -> tuple[NDArray[np.int64], NDArray[np.float64], NDArray[np.str_]]:
-    """Load data from a CSV file from the specified data_module.
+    """Load data from a CSV file from the specified module_name.
 
     Parameters
     ----------
     file_name: str, required
-        The filename of the CSV file to load from `data_module/file_name`.
-    data_module: str or module, default='sknnr.datasets.data'
+        The filename of the CSV file to load from `module_name/file_name`.
+    module_name: str or module, default='sknnr.datasets.data'
         The module where the data file is located.
 
     Returns
@@ -93,7 +93,7 @@ def load_csv_data(
     The CSV must be formatted with plot IDs in the first column and data values in the
     remaining columns. The first row must contain the column names.
     """
-    with _open_text(data_module, file_name) as csv_file:
+    with _open_text(module_name, file_name) as csv_file:
         data_file = csv.reader(csv_file)
         headers = next(data_file)
         rows = list(iter(data_file))
@@ -111,7 +111,7 @@ def load_dataset_from_csv_filenames(
     target_filename: str,
     return_X_y: bool = False,
     as_frame: bool = False,
-    data_module: str | types.ModuleType = DATA_MODULE,
+    module_name: str | types.ModuleType = DATA_MODULE,
 ) -> tuple[NDArray[np.float64], NDArray[np.float64]] | Dataset:
     """Load separate data and target CSV files into a dataset or paired NumPy arrays.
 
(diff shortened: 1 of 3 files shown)


grovduck:
@aazuspan, at long last, here's a first cut at exposing those two functions for use in other `sknnr` related projects.  I think the code changes are more or less OK, but could use your help with knowing whether the tests are sufficient.

On testing, it looks like `scikit-learn`'s `r2_score` function [now defaults](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.r2_score.html) to `multioutput="uniform_average"` which is an average across all outputs.  If we use `multioutput="raw_values"`, the output is correct.  However, I didn't want to bury this commit in this PR in case you thought it was better handled as a separate PR that we push in front of this one.  For now, I'll leave the tests as failing.

src/sknnr/datasets/_base.py
```diff
@@ -66,17 +66,35 @@ def _open_text(module_name: str, file_name: str) -> IO[str]:
     return resources.open_text(module_name, file_name)
 
 
-def _load_csv_data(
+def load_csv_data(
     file_name: str,
+    *,
+    data_module: str = DATA_MODULE,
```
aazuspan, line 72: I should've done this originally, but this (and `_open_text` and `load_dataset_from_csv_filenames`) should probably be annotated with `str | types.ModuleType`.

aazuspan (commented):
@grovduck, looks great! I had one nit on type annotations, but the tests look sufficient to me, since we're already testing these indirectly with our other dataset tests.

> On testing, we're failing on an unrelated issue. It looks like scikit-learn's r2_score function [now defaults](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.r2_score.html) to multioutput="uniform_average" which is an average across all outputs. If we use multioutput="raw_values", the output is correct. However, I didn't want to bury this commit in this PR in case you thought it was better handled as a separate PR that we push in front of this one. For now, I'll leave the tests as failing.

Good find! I'm fine if you want to include that fix here for simplicity, or as a separate PR.

grovduck:
> On testing, we're failing on an unrelated issue. It looks like `scikit-learn`'s `r2_score` function [now defaults](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.r2_score.html) to `multioutput="uniform_average"` which is an average across all outputs. If we use `multioutput="raw_values"`, the output is correct. However, I didn't want to bury this commit in this PR in case you thought it was better handled as a separate PR that we push in front of this one. For now, I'll leave the tests as failing.

I decided to address this in a separate PR (#70) and my reasoning for why the test was failing was flawed, but explained in that issue.  I've now merged those changes into `expose-dataset-functions` branch.

grovduck:
Thanks @aazuspan.  One last nit on naming.  The kwarg for the module name is called `module_name` in `open_text` and `data_module` in both `load_csv_data` and `load_dataset_from_csv_filenames`.  Probably good to keep them consistent.  I had used `data_module` to be consistent with [`sklearn.datasets.load_csv_data`](https://github.com/scikit-learn/scikit-learn/blob/c44457af96a924683b6a65bd3fd6c285043bdec6/sklearn/datasets/_base.py#L318), but I think internal consistency would be better.

aazuspan:
I agree that keeping that name consistent makes sense. I'm happy with whatever you decide to go with.

grovduck:
Cool, I stuck with `module_name`.  I'll go ahead and merge.

grovduck:
@aazuspan, just realized that if I'm going to use this in `synthetic-knn`, we're going to need to push a release (unless I rely on a editable version of `sknnr`)?  I'm out of my depth on whether or not it makes sense to put out a new release and what the naming convention might be.  I thought you might have a bit more insight here.

aazuspan:
Good point - probably no reason not to push another alpha release. Our naming convention is a little inconsistent right now since our tag is `0.1.0-alpha` and our PyPI release is `0.1.0a0`. I think the former is the correct semver protocol, but `hatch` generated the latter. I'd suggest we stick with the former and increment to `0.1.0-alpha.1`. I think the steps will be as described in the [contributing guide](https://sknnr.readthedocs.io/en/latest/contributing/#releasing), with the caveats of 1) we'll need to manually set the version since `hatch` doesn't support alpha releases and 2) we'll probably need to manually rename the files generated by `hatch build` to avoid pushing the weird version name that `hatch` creates to PyPI.

I'm happy to push the release, unless you'd rather try your hand at it.

grovduck:
> I'm happy to push the release, unless you'd rather try your hand at it

If you don't mind, I'll have you do this release and I'll look at your changes so that I can try next time.  Your directions in the contributing guide are very clear, but I'm still a bit cautious!

aazuspan:
After only a little confusion, `0.1.0a1` is now [live on PyPI](https://pypi.org/project/sknnr/0.1.0a1/)!

As you might have followed in the PRs, I tried using the semver alpha specifier of `0.1.0-alpha.1`, but realized too late that PyPI won't accept that and requires `0.1.0a1` instead. Rather than keep things inconsistent between our package version in `__about__.py`, our Git tag, and our PyPI release, I decided to just switch them all over to the PEP440 style of `0.1.0a1`.

I also went back and renamed the original Git tag to `0.1.0a0` for consistency with the PyPI release. I think a `git pull` should put you in sync with the tags, but it's possible you'll have a leftover `0.1.0-alpha` tag locally.

grovduck:
Oh man, I'm glad you dealt with this one! :wink:  I'm sure I'd still be trying to figure it out!  Thanks for the heads up on the leftover tag - I found it and deleted it.  

Correct me if I'm wrong, but it looks like `hatch version a` will update the version from `0.1.0a1` to `0.1.0a2`?  Did you use that when updating?  I think we can update the contributing guide if you think this will work.

aazuspan:
> Correct me if I'm wrong, but it looks like hatch version a will update the version from 0.1.0a1 to 0.1.0a2? Did you use that when updating? I think we can update the contributing guide if you think this will work.

Great find! I didn't realize `hatch` supported anything other than major/minor/patch, but you're right. I'll update the contributing guide to include that.

PR #70, opened 2024-07-10
Fix `r2_score` port test

grovduck:
As part of using the [Array API](https://github.com/scikit-learn/scikit-learn/pull/27904) for `r2_score`, the return value of `r2_score` is now cast explicitly as a `float` rather than `np.float64` when it is a scalar value.

From [this doc](https://github.com/scikit-learn/scikit-learn/blob/c44457af96a924683b6a65bd3fd6c285043bdec6/doc/modules/array_api.rst?plain=1#L138-L152):

> Note however that scoring functions that return scalar values return Python scalars (typically a `float` instance) instead of an array scalar value.

In `test_port.test_score_independent`, we were calling `r2_score` using the default `multioutput="uniform_average"` which was returning a single `np.float64`.  We were then mistakenly calling `.mean()` on this value which had not previously thrown an error because it was operating on a `np.float64`.  With the change in return type, this test now throws an error at `scikit-learn>=1.5`.

Simply removing `.mean()` corrects this test.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -26,6 +26,9 @@ jobs:
           python -m pip install --upgrade pip
           pip install hatch
 
+      - name: List dependencies
+        run: hatch run pip freeze
+
       - name: Run tests
         run: hatch run test:all
   lint:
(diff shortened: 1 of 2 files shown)


grovduck:
@aazuspan, my reasoning in #69 on this error was flawed and I thought it would be a good idea to keep this separate.  I've also added a slight change to the CI config to explicitly list the dependency package versions used in the CI tests.  There might be a way to get this info from the CI logs already, but I coudn't find it.

aazuspan (approved):
Interesting, thanks for explaining the underlying issue. Makes sense! The `pip freeze` is a good addition.

LGTM!

PR #73, opened 2024-07-11
Update `hatch version` guide

aazuspan:
As pointed out by @grovduck, `hatch version` supports release types other than major/minor/patch.

### DIFF
--- a/docs/pages/contributing.md
+++ b/docs/pages/contributing.md
@@ -76,10 +76,10 @@ Whenever the docs are updated, they will be automatically rebuilt and deployed b
 
 ### Releasing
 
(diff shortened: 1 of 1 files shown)


grovduck (approved):
@aazuspan, this looks great.  Good to have the reminder to isolate the fix into a new release branch and merge into `main`.  I know you created a PR for the release branch (presumably so that I could follow along), but is the general advice to just go ahead and do the merge without a PR?

I think the link to hatch answers any questions a contributor might have and we don't need to provide additional detail here.

aazuspan:
> but is the general advice to just go ahead and do the merge without a PR?

Yeah, that's probably fine, since CI isn't likely to catch any issues with a simple version bump.

Thanks for taking a look!

PR #76, opened 2024-07-26
Add Hatch script for local Python matrix tests

### LINKED ISSUE #64: Test all Python versions with Hatch 1.8.0
The new [Hatch release](https://github.com/pypa/hatch/releases/tag/hatch-v1.8.0) will auto-install missing Python versions when using an [environment matrix](https://hatch.pypa.io/latest/environment/#matrix) for testing! I did a quick test by adding:

```toml
[[tool.hatch.envs.test.matrix]]
python = ["3.8", "3.9", "3.10", "3.11", "3.12"]
```

to the `pyproject.toml`, updating my `hatch`, and running `hatch run test:all`, and it successfully ran the whole test matrix (takes a while the first time, but it looks like it caches installs for future runs). This solves all the trouble I've had in the past getting `tox` to find the right Python installations.

@grovduck I don't think this is a high priority, or something we necessarily even want to run every time, but it would be nice to have the option to test old Python versions and catch those sneaky typing incompatibilities before they hit the CI. I figured you might be interested in this for other projects that you've migrated to Hatch as well.

aazuspan:
Closes #64 by adding a `test_matrix` Hatch env for testing across Python versions (e.g. `hatch run test_matrix:all`).

@grovduck, I just copied the config from `synthetic-knn` to keep everything consistent, and will duplicate this in `sknnr-spatial` next.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,12 @@ dependencies = ["pytest", "pytest-cov", "pandas"]
 all = "pytest . {args} --doctest-modules"
 coverage = "pytest . --cov=src/sknnr {args} --doctest-modules"
 
+[tool.hatch.envs.test_matrix]
+template = "test"
+
+[[tool.hatch.envs.test_matrix.matrix]]
+python = ["3.9", "3.10", "3.11", "3.12"]
+
 [tool.hatch.envs.docs]
 dependencies = ["mkdocs", "mkdocs-material", "mkdocstrings[python]"]
 


grovduck (commented):
@aazuspan, I've confirmed that `hatch run test_matrix:all` works for me.

This might be a dumb question, but it looks like we are still `requires-python = ">=3.8"` on this project.  Do we need to add `3.8` to the matrix or are we ready to drop `3.8` support?  Looks like `3.8` is [end-of-life](https://devguide.python.org/versions/) as of October 2024 (and `3.13` is just on the horizon!)

aazuspan:
Good catch, I forgot we were still supporting it! In the interest of getting all the `sknnr` packages in sync, I'm in favor of dropping `3.8` now. If that sounds good to you, I'll work on getting that merged first and then come back to this.

grovduck:
> Good catch, I forgot we were still supporting it! In the interest of getting all the `sknnr` packages in sync, I'm in favor of dropping `3.8` now. If that sounds good to you, I'll work on getting that merged first and then come back to this.

Sounds good to me.

PR #77, opened 2024-08-05
Drop Python 3.8 support

aazuspan:
Raises the minimum supported Python to 3.9, in anticipation of end-of-life for 3.8 in a couple months.

That allowed us to remove an internal compatibility function and use the "new" parenthesized `with` block. Unrelated, I removed the `target-version` specifier to allow Ruff to infer it from `requires-python` and added the `dist` build directory to `.gitignore`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,7 +8,7 @@ jobs:
   test:
     strategy:
       matrix:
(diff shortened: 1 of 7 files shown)


grovduck (approved):
@aazuspan, nothing to add on this one.  Looks nice and clean.  Good catch on getting rid of `_open_text`.  I likely would have forgotten about that one!

PR #80, opened 2024-12-09
Tag fixes for `sklearn==1.6.0`

aazuspan:
There were a few more breaking changes between `sklearn==1.6.0rc1` and `1.6.0` (now released) which this resolves:

1. `multi_label` is no longer an attribute of `RegressorTags` (I've tried to find where this was changed to figure out the rationale, but didn't have any luck. Maybe it was never implemented but it just wasn't being checked before?).
2. `check_positive_only_tag_during_fit` was added, which uses 1D data that breaks CCA.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ include = ["/src"]
 dependencies = ["pre-commit"]
 
 [tool.hatch.envs.test]
-dependencies = ["scikit-learn>=1.6.0rc1", "pytest", "pytest-cov", "pandas"]
+dependencies = ["scikit-learn>=1.6.0", "pytest", "pytest-cov", "pandas"]
 
 [tool.hatch.envs.test.scripts]
 all = "pytest . {args} --doctest-modules"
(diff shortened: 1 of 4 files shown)


grovduck (approved):
@aazuspan, sorry so slow on review!  I think you were hoping that this one gets merged first so that you can make another release for `sknnr-spatial`?  [**EDIT:** sorry, I read that incorrectly in the `sknnr-spatial` issue.  You are making that test more generic now.]

I don't have any qualms about merging this one in, but I am getting a warning (in `_ccora.py`) when running locally that I'm not seeing in the CI tests.  Just thought I'd mention it in case it triggers anything for you.

warning

There is also the weird warning about not being able to delete the temp test directory as well.

aazuspan:
Thanks Matt! I'm not seeing those warnings on my end - maybe a version mismatch between our environments? Neither looks too worrying as long as we're not filling up your hard drive with temp test data though, so I'll go ahead and merge and we can revisit later if they crop up again.

PR #81, opened 2024-12-13
Bump `0.1.0a1` -> `0.1.0a2`

aazuspan:
@grovduck, I'm gonna make another quick alpha release to get a version on PyPI that's compatible with `sklearn>=1.6`. I've been working on some documentation for `sknnr-spatial`, and keep having to work around those breaking changes.


grovduck:
@aazuspan, can you check the SHA of the `0.1.0a2` tag?  To me it looks like it is one commit behind (ca383c5) where we did this PR (11b649d).  It's entirely possible that I'm getting that wrong.

aazuspan:
You're right, I tagged `0.1.0a2` before updating the version (https://github.com/lemma-osu/sknnr/blob/0.1.0a2/src/sknnr/__about__.py). My mistake! I checked and the PyPI release is correct at least.

Does it make sense to force push a new `0.1.0a2` tag at the right commit?

grovduck:
> Does it make sense to force push a new `0.1.0a2` tag at the right commit?

Yep, I think that sounds safe!

aazuspan:
The tag should be correct now, but somehow I seem to have [broken an estimator check](https://github.com/lemma-osu/sknnr/actions/runs/12819328112/job/35746752799) in the process. Looks like the last tests ran with `scikit-learn==1.6.0` and it's now running with `scikit-learn==1.6.1`, so I'm betting that's the culprit... I'll open an issue to track. Hopefully that doesn't block you!

Issue #83, opened 2025-01-16
Fix failing tests caused by inherited sparse tag in `scikit-learn==1.6.1`


aazuspan:
Tests began failing between the [previous CI run](https://github.com/lemma-osu/sknnr/actions/runs/12322168413) and the [latest CI run](https://github.com/lemma-osu/sknnr/actions/runs/12819328112/job/35746753223) with no code changes. Looking at the dependencies, I'm betting it's due to a switch from `scikit-learn==1.6.0` to `scikit-learn==1.6.1`. 

~There's nothing in [the changelog](https://scikit-learn.org/dev/whats_new/v1.6.html#version-1-6-1), but it seems like it's running a sparse test when we don't support sparse data.~ Example error below:

```
FAILED tests/test_estimators.py::test_sklearn_estimator_checks[EuclideanKNNRegressor()-check_estimator_sparse_tag] - AssertionError: Estimator EuclideanKNNRegressor raised an exception. The tag self.input_tags.sparse=True might not be consistent with the estimator's ability to handle sparse data (i.e. controlled by the parameter `accept_sparse` in `validate_data` or `check_array` functions).
```

EDIT: Oops, there's definitely something in the changelog! From https://github.com/scikit-learn/scikit-learn/pull/30187:

> if input_tags.sparse=False the estimator must raise a error when fitted on sparse data

aazuspan:
@grovduck, I think I got this figured out. As of 1.6.1, we now inherit `input_tags.sparse = True` [from `NeighborsBase`](https://github.com/antoinebaker/scikit-learn/blob/b54a408c0bad6dd4eaa5329cf29acbec3e43bdb9/sklearn/neighbors/_base.py#L710), but we raise an error for sparse data in our `validate_data` checks because `accept_sparse` defaults to `False`. That breaks the newly added check since those are incompatible.

The easy fix is to override `input_tags.sparse = False`. The harder fix is to figure out whether we actually do support sparse data. I'm okay kicking that can down the road, but what do you think?

grovduck:
@aazuspan, like you say, at this point I'm not confident to know whether we support sparse data for our estimators and whether this would be different for `X` and `y` matrices (is that even an option?).  So I agree that we override `input_tags.sparse = False` for now and revisit later.

aazuspan:
Great, I can tackle this and #82 unless you've got something in progress?

grovduck:
[...]

PR #84, opened 2025-01-17
Finalize estimator tags and fix sparse test failure

### LINKED ISSUE #82: Finalize estimator tags
## Context

`scikit-learn==1.6.0` overhauled the [estimator tags](https://scikit-learn.org/dev/developers/develop.html#estimator-tags). #79 and #80 refactored our estimators to be compatible with the new system, which raised some good discussion on the accuracy of our current tags

## Current Tags

I was struggling to keep track of which tags are defined and/or inherited, so I made the figure below to show the boolean tags (excludes `estimator_type` and `transformer_tags.preserves_dtype`) for all our estimators (gist [here](https://gist.github.com/aazuspan/c0f2f51fe3f37a6c301437e5ca1dc493) if we need to recreate it).

yellow=`True`, purple=`False`, white=`NaN`

plot_tags

## Tags

We can expand this as we go, but here are the things that I think we've already discussed or that stick out to me. Please correct me anywhere I'm wrong @grovduck.

### `estimator_type`

Not included in the figure above, but our regressors are all listed as `regressor` while our transformers are `None`. This is consistent with `sklearn` transformers, but seems wrong since `transformer` [is listed](https://scikit-learn.org/dev/modules/generated/sklearn.utils.Tags.html) as an option. 

There's probably no benefit to adding this, but I don't think it would hurt either.

### `target_tags.single_output`

See https://github.com/lemma-osu/sknnr/pull/79#discussion_r1882609450. I think where we've settled is that `MSNRegressor` and `CCorATransformer` *should* support single output as 1) it is technically possible, 2) there are valid applications e.g. single-species models, 3) it is supported by `yaImpute`. No change required there.

`GNNRegressor` and `CCATransformer` do *not* support single output, so those should be explicitly set to `False`.

### `target_tags.required`

I believe this should be `True` for CCA and CCorA, based on their strict requirements. Currently it's only `True` for CCA.

### `target_tags.one_d_labels` and `target_tags.two_d_labels`

These are [defined as](https://scikit-learn.org/stable/modules/generated/sklearn.utils.TargetTags.html): 

> Whether the input is a 1D labels (y).

and 

> Whether the input is a 2D labels (y).

I find those a little confusing (grammar aside), but I would have assumed that they correspond with `input_tags.one_d_array` and `input_tags.two_d_array` and represent support for `y` array shapes, since "label" is sometimes used interchangeably to refer to dependent/target variables. However, the only place I saw this set to `True` in the `sklearn` code base was in `LabelEncoder` and `LabelBinarizer`, which makes me think they're using a more specific definition of "label".

aazuspan:
Closes #82 and #83

- CCA now correctly requires positive targets
- CCorA now correctly requires targets
- All estimators (other than `RawKNNRegressor`) now explicitly prohibit sparse data. This is a temporary measure to fix failing checks. A closer investigation of sparse support is needed.
- The addition of `target_tags.required` to CCorA triggered an estimator check `check_requires_y_none` which looks for an expected error message when passed `y=None`. That error is now correctly raised by passing y into `_validate_data`. This removed the need to separately check the y array. Note that we don't mutate `X` with the validation check as this will strip feature names that must be passed to `StandardScalerWithDOF`.

@grovduck, please check and confirm that I covered all the tag changes we discussed, thanks!

The updated boolean tags following these changes:

image

### DIFF
--- a/src/sknnr/_base.py
+++ b/src/sknnr/_base.py
@@ -149,6 +149,12 @@ class TransformedKNeighborsRegressor(RawKNNRegressor, ABC):
             return_dataframe_index=return_dataframe_index,
         )
 
+    def __sklearn_tags__(self):
+        tags = super().__sklearn_tags__()
+        tags.input_tags.sparse = False
+
+        return tags
+
 
 class OrdinationKNeighborsRegressor(TransformedKNeighborsRegressor, ABC):
     """
--- a/src/sknnr/transformers/_cca_transformer.py
+++ b/src/sknnr/transformers/_cca_transformer.py
@@ -47,5 +47,6 @@ class CCATransformer(ComponentReducerMixin, TransformerMixin, BaseEstimator):
     def __sklearn_tags__(self):
         tags = super().__sklearn_tags__()
         tags.target_tags.required = True
+        tags.target_tags.positive_only = True
 
         return tags
--- a/src/sknnr/transformers/_ccora_transformer.py
+++ b/src/sknnr/transformers/_ccora_transformer.py
@@ -1,6 +1,5 @@
-import numpy as np
 from sklearn.base import BaseEstimator, TransformerMixin
-from sklearn.utils.validation import check_array, check_is_fitted
+from sklearn.utils.validation import check_is_fitted
 
 from .._base import _validate_data
 from . import ComponentReducerMixin, StandardScalerWithDOF
@@ -9,10 +8,9 @@ from ._ccora import CCorA
 
 class CCorATransformer(ComponentReducerMixin, TransformerMixin, BaseEstimator):
     def fit(self, X, y):
-        _validate_data(self, X=X, reset=True)
+        _, y = _validate_data(self, X=X, y=y, reset=True, multi_output=True)
         self.scaler_ = StandardScalerWithDOF(ddof=1).fit(X)
 
-        y = check_array(y, input_name="Y", ensure_2d=False, dtype=np.float64)
         if y.ndim == 1:
             y = y.reshape(-1, 1)
         y = StandardScalerWithDOF(ddof=1).fit_transform(y)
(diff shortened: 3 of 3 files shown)


src/sknnr/_base.py
```diff
@@ -149,6 +149,12 @@ def kneighbors(
             return_dataframe_index=return_dataframe_index,
         )
 
+    def __sklearn_tags__(self):
```
grovduck, line 152: Stupid question, but why did you set `tags.input_tags.sparse` here instead of in `RawKNNRegressor`.  I notice in your matrix that `sparse = True` for `RawKNNRegressor`.  Is that intentional?

grovduck (approved):
Looks good to me!  One small question, but otherwise looks ready to merge.

src/sknnr/_base.py
aazuspan: My rationale was that the minor differences between `RawKNNRegressor` and `KNeighborsRegressor` shouldn't affect sparse data support, so I was pretty confident inheriting that without any further investigation (beyond passing estimator checks). What do you think? Better to set everything `False` to start with?

grovduck: That's kind of what I figured.  We might want to slightly edit the top post in this PR to be:

> * All estimators now explicitly prohibit sparse data (other than `RawKNNRegressor)

I'm fine with keeping what you have as-is.

grovduck:
@aazuspan, looks like we can close #83 as well?

grovduck:
@aazuspan, for some reason #83 still looks to be open?  I think I [just learned](https://github.com/nus-cs2103-AY2021S2/forum/issues/293) that you have to use "closes #82, closes #83" and that the "and" doesn't work?  I may be off, though ...

aazuspan:
Thanks for the heads up! I just closed manually, but good to know about linking multiple issues :)

Issue #87, opened 2025-03-28
RFNNRegressor should support both classifier and regressor RFs when building forests


grovduck:
In #85, we created the `RFNNRegressor` estimator that uses a suite of random forests (one for each attribute in `y` (or `y_fit`)) to support a kNN approach to match targets to reference data based on similarity in the node IDs across forests and their trees.  In that PR, we relied exclusively on `sklearn.ensemble.RandomForestRegressor` to build the individual forests, assuming that all `y` features were continuous rather than categorical.  We decided to defer the implementation of using *both* `sklearn.ensemble.RandomForestRegressor` and `sklearn.ensemble.RandomForestClassifier` to build the forests for a later time.  Per [this comment](https://github.com/lemma-osu/sknnr/pull/85#issuecomment-2628508097):

> The other obvious thing I need to do to make this estimator more useful is to have the set of random forests to be both `RandomForestClassifier` and `RandomForestRegressor` objects. The `RandomForestClassifier` forests could either be built from classifying continuous `y` attributes into classes or from `y` features that already represent classes, either numeric or strings. In my mind, this seems like we need to both: 1) tell the estimator how to build each forest (some kind of mapper between y attribute and random forest type); and 2) have utility functions to bin continuous data into categorical data. I can definitely see the argument that 2. should be outside the scope of the estimator such that the user does this on their own, but it might be nice to have example utility functions to show how this is done.

### Design considerations
- We need to decide what feature dtypes will we support with `RFNNRegressor`.  Ideally, we would be able to support `np.object` (string) and `pd.Categorical` features and well as integer and floating-point features (I suppose we could also support `np.bool`, but building a forest with just two classes seems pretty limited).  Both pandas dataframes and numpy arrays can handle mixed data types, so passing a `y` frame or array should be possible. (Note that, at present, our test datasets only support floating-point features - this isn't a show-stopper, but we'll need to generate mixed-type dataframes if we want to support other data types).
[...]

PR #90, opened 2025-04-15
Change default value for max_features to match scikit-learn

grovduck:
When first implementing #85, we wanted to ensure that we were matching yaImpute's RFNN implementation as closely as possible.  The default value for R's randomForest `mtry` parameter is `sqrt`, presumably because the assumption is that it is run most often in classification mode.  For our initial implementation of RFNN, we used scikit-learn's `RandomForestRegressor` but failed to recognize that the default value for `max_features` (the scikit-learn corollary to `mtry`) is actually 1.0 rather than `sqrt`.  

This PR sets the correct default value for both `RFNNRegressor` and `RFNodeTransformer` to 1.0.  As we now have regression test data in place, this also necessitated `--force-regen` on the 8 files associated with the `RFNNRegression` estimator.

**EDIT**: @aazuspan rightly questions whether this would change our original results from `yaImpute`.  Unfortunately, as part of the original port and detailed in [this comment](https://github.com/lemma-osu/sknnr/pull/85#issuecomment-2613658252), we weren't able to replicate `yaImpute`'s initial results based on differences with the fundamental random forest implementations in R and scikit-learn.  So, the results that we replaced with regression tests in #89 were already different from the `yaImpute` results.


grovduck:
@aazuspan, hopefully this is a quick and dirty one.  I wanted to keep it separate from any work on #87, but this change will help fix some of my failing tests there.

aazuspan (approved):
Good catch, LGTM!

aazuspan:
Actually, on second thought, would it make sense to keep the tests using `sqrt` so that the results remain equivalent with `yaimpute`? I'm not sure how much we want to try to preserve the original port test results.

grovduck:
> Actually, on second thought, would it make sense to keep the tests using `sqrt` so that the results remain equivalent with `yaimpute`? I'm not sure how much we want to try to preserve the original port test results.

Yep, great question and I neglected to mention this in the initial comment.  I've now edited that comment to address this.  I think I'd rather be consistent with `scikit-learn` estimators from this point forward.

aazuspan:
That makes sense, thanks for the clarification!

PR #93, opened 2025-05-02
Add PyPI publish workflow

aazuspan:
This adds a workflow to automate publishing to PyPI whenever a new tag is pushed and updates the contributing guide with the relevant instructions.

See https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/.

@grovduck, I've added Github as a trusted publisher [on PyPI](https://pypi.org/manage/project/sknnr/settings/publishing/), so this should work next time that we need to make a release. Unfortunately, I don't think there's any good way to validate the workflow, but I'm reasonably confident it'll work as expected since I copied it directly from the [`eerepr` workflow](https://github.com/aazuspan/eerepr/actions/workflows/publish.yml) which ran successfully.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,0 +1,56 @@
+# Adapted from: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
+name: Publish to PyPI
+on:
+    push:
+        tags:
+            # Publish when a tag is pushed
+            - '*'
+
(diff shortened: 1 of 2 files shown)


grovduck (approved):
@aazuspan, I'm not too qualified to review the GA workflow, but totally trust that it will work because it's working on `eerepr`.  Looks like the only change is the PyPI URL.

The edits to the contributing page look great as well.  Much easier!

PR #95, opened 2025-05-20
Run CI tests in parallel

### LINKED ISSUE #94: Parallelize testing
The full test set is getting pretty slow. I did a quick experiment using [pytest-xdist](https://github.com/pytest-dev/pytest-xdist) to parallelize testing and it decreased the time for the full test set from 217s to 48s with 6 workers on my laptop.

```
hatch -e test run pip install pytest-xdist
hatch run test:all -n auto
```

@grovduck, do you want to give this a try and make sure it works well with your setup? If so, I'm not sure there's any downside aside from an extra test dependency.

aazuspan:
Closes #94 by adding a test dependency on `pytest-xdist` and parallelizing CI tests. CI uses all 4 available threads (per [Github docs](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories)), which decreased overall CI time from about 3.5 minutes to 2.25, based on just one run.

Local tests are still serial by default but can be run in parallel with (using 8 threads):

```
hatch run test:all -n 8
```

@grovduck, we could parallelize locally by default (i.e. `hatch run test:all` would always use `-n auto`), but I don't have strong feelings either way. As you mentioned in #94, using `auto` adds some additional overhead so we might always end up overriding it manually anyways. Let me know if you have a preference.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -30,7 +30,7 @@ jobs:
         run: hatch run pip freeze
 
       - name: Run tests
-        run: hatch run test:all
+        run: hatch run test:all -n 4
   lint:
     runs-on: ubuntu-latest
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,6 +36,7 @@ dependencies = [
     "pytest",
     "pytest-cov",
     "pytest-regressions",
+    "pytest-xdist",
     "pandas",
 ]
 


grovduck (approved):
@aazuspan, LGTM.

> @grovduck, we could parallelize locally by default (i.e. hatch run test:all would always use -n auto), but I don't have strong feelings either way. As you mentioned in https://github.com/lemma-osu/sknnr/issues/94, using auto adds some additional overhead so we might always end up overriding it manually anyways. Let me know if you have a preference.

I think I'd prefer to NOT parallelize by default locally based on what I found in #94 or possibly to use fewer cores to do so.  Another option(?) is to create a separate script for `tool.hatch.envs.test.scripts`:
```
parallel = "pytest . {args} --doctest-modules -n auto"
```
but that might be overkill.

PR #99, opened 2025-08-29
Implementation of Gradient Boosting Nearest Neighbors (`GBNN`)

### LINKED ISSUE #96: Use `GradientBoosting` in a `KNearestNeighbors` context
@aazuspan, I know you've brought this one up and I've thought about it before, so I think it's time to start collecting our thoughts as to whether we should try to add a gradient boosting nearest neighbor (GBNN) to `sknnr`.  Given that we have most of the hard work out of the way on `RFNNRegressor` and that there is a lot of overlap between the two methods (random forest and gradient boosting), it seems like an optimal time to think about this, especially if we're able to reuse/refactor some of the `RFNodeTransformer` code.

My understanding of gradient boosting isn't top-notch, but generally I understand that shallow trees (weak-learners) are built based on regression targets and each subsequent tree is based on targets that represent the residuals from the previous step.  Predictions are based on the additive results on each weak learner, e.g. `prediction = tree_1(x) + tree_2(x) + ... + tree_n(x)`.  In a classification context with multi-classes (rather than just binary), each class' estimators are evaluated and then the class with the highest score (softmax) is chosen.

As with `RandomForestClassifier` and `RandomForestRegressor`, the `GradientBoostingClassifier` and `GradientBoostingRegressor` classes provide the `apply` method which provides the node IDs from each weak learner, which is fundamental for building the node matrix.  However, as opposed to random forest where each tree can be thought of as independent and has equal weighting in prediction, I'm having a hard time wrapping my head around whether gradient boosting trees can be thought of in the same way.  Because the trees are a series that rely on the residuals of the previous step, should each node carry equal weight in prediction.  My intuition leads me to think that it's more important to be "right" early on in the series where the residuals would presumably be larger.  I'm trying to imagine a scenario where a target sample has roughly equal nodes shared with two reference samples, but the shared nodes with the first sample happen at the beginning of the set of trees and the shared nodes with the second sample happen at the end of the set of trees.  Is one more correct than the other?  Another way of saying this is whether predictions are more accurate if a reference sample "derails" from its ideal path early or late in the sequence.

I don't think it would be difficult to introduce a `tree_weights` parameter that would modify the Hamming distance and by default it would be set to equal weighting.   This would give us the opportunity to explore these questions without blind intuition.  BUT, I'm interested if you have more informed thoughts about how gradient boosting might behave in this context and whether we need to do something different than just simple Hamming distance.

Also, if I've completely missed the boat in terms of how you think this might be implemented, I would love to hear those thoughts!

grovduck:
This PR adds support for using ensembles of [`GradientBoostingRegressor`](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.GradientBoostingRegressor.html) and [`GradientBoostingClassifier`](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.GradientBoostingClassifier.html#sklearn.ensemble.GradientBoostingClassifier) models in a kNN context.  As far as we're aware, this is the first implementation of using gradient-boosting models paired with nearest neighbor imputation, but is a natural extension of `RFNN` first developed by Crookston and Finley and implemented in `sknnr`.

Separate gradient-boosting estimators are built for each target feature present in `y` (or `y_fit`) and the type of estimator (either regression or classification) is determined by the data type of each target (floating-point and integer use regression estimators, string and `pd.Categorical` use classification estimators).  As with RFNN, node IDs from the reference samples that were used to fit the model are captured.  New samples are run through the fit estimators and their node IDs are compared to reference node IDs using a weighted Hamming distance to identify nearest neighbors.

Our intuition is that the individual (and successive) trees in each gradient-boosting model should have separate weights associated with them.  @aazuspan [first suggested](https://github.com/lemma-osu/sknnr/issues/96#issuecomment-2967847215) weighting trees in an estimator proportional to the change in loss that is explained with each step.  This is the default option for the `tree_weighting_method` parameter for `GBNNRegressor`, although `uniform` weighting is also available as an option.

@aazuspan, maybe a lot to wade through on this PR, but a few specific areas of interest/concern:

- The tree weighting algorithm that you suggested is the bulk of the new code here.  I decided to split that off into a separate function (as well as `uniform` weighting which can be used by both RF and GB).  I've called your method `delta_loss`, but open to different names as well.  What are your thoughts on accepting a function as a parameter to `tree_weighting_method`?
- In testing (specifically on `test_estimators::test_gridsearchcv`), I ran into the situation where the GB models failed to fit.  I don't think `GridSearchCV` fails on these fits, but instead the values for `est.train_score_` seem to all be set to 0.0.  This was causing the normalizing step (e.g. `loss_delta / np.sum(loss_delta)`) to fail, so for now I'm just putting in equal weights when this condition happens.  My thought is that the model fitting would raise an exception, so it wouldn't typically get to this step, but that's not a good solution for users that want to use it in a grid search context.
- I didn't write any new transformer tests other than to verify that the regression/classifier keywords get passed appropriately, but it feels like we need to do some checking on values coming out of `delta_loss`.  We can't test on monotonicity based on your example, but is there anything we should be testing on for "correctness" here?
- Similar to that, I also haven't handled the multi-classification weighting that we raised [here](https://github.com/lemma-osu/sknnr/issues/96#issuecomment-2970983522).  I'll continue to work on that.
- As of now, I am using the squared-loss delta, but you mentioned that it might be better to sqrt that when calculating deltas.  For that, we'd also need to factor in the `loss` parameter as you noted: "Generalizing that to work with other loss functions, and figuring out whether it actually works in practice would take some more effort."

Closes #96


grovduck:
> Similar to that, I also haven't handled the multi-classification weighting that we raised [here](https://github.com/lemma-osu/sknnr/issues/96#issuecomment-2970983522).  I'll continue to work on that.

Currently, this won't even work - slipped through the tests as I wasn't testing with multi-class data.
```
from sklearn.datasets import make_classification

from sknnr import GBNNRegressor
from sknnr.transformers import GBNodeTransformer

# Create a sample dataset with three classes
X, y = make_classification(
    n_samples=200,
    n_features=10,
    n_informative=5,
    n_redundant=2,
    n_classes=3,
    n_clusters_per_class=1,
    random_state=42,
)

# Reclassify `y` as string to force `GradientBoostingClassifier`
y = np.reshape(y.astype(str), (-1, 1))

# Create a GBNodeTransformer
transformer = GBNodeTransformer().fit(X, y)
print(transformer.transform(X).shape) # Returns (200, 100, 3) - need it to return (200, 300)

# Create a GBNNRegressor
est = GBNNRegressor().fit(X, y)
```
This errors with:
```
ValueError: Found array with dim 3, while dim <= 2 is required by GBNNRegressor.
```
Fixing the node indices shouldn't be too bad, we will just need to reshape when an estimator has `n_classes_ > 1`.  And the current `delta_loss` actually works in contexts with `n_classes_ > 1` as `train_score_` returns an array of shape `(n_estimators_, 1)`.  So if we simply broadcast that to a shape of `(n_estimators_, n_classes_)`, that would be identical weights for each `DecsionTreeRegressor` (one per class) for each tree level, e.g.
```
[[0.57 0.57 0.57]
 [0.42 0.42 0.42]
 [0.32 0.32 0.32]
 ...
]
```
Whether or not this is what we want is a different story.  We would then reshape to match the node indexes shape.

aazuspan:
@grovduck, this looks awesome! I'm a little hung up on some of the questions around converting losses to weights, so these are just some quick responses on that topic - I still need to give everything else a proper read.

> I've called your method delta_loss, but open to different names as well.

I like `delta_loss` on its own, but thinking about other potential weighting methods we might add in the future, I wonder if we might want to have an option that uses the `oob_improvement_` attribute? If so, maybe we'd want something like `train_improvement` to match that?

> As of now, I am using the squared-loss delta, but you mentioned that it might be better to sqrt that when calculating deltas.

In hindsight, I think trying to make automatic rules about converting losses to weights (e.g. square-root transforming some metrics but not others) seems like it opens a can of worms for us. I'm not sure if I like this suggestion, but maybe we add another weighting method `sqrt_delta_loss` / `sqrt_train_improvement`? I guess we could also add another parameter like `loss_to_weight_func` where you could pass a callable like `np.sqrt` to make it more flexible, but I don't love that either.

The simplest solution (which we could revisit later) would probably just be to clearly document that the losses are mapped directly to weights, so choose your loss function accordingly (i.e., maybe use `absolute_error` instead of `squared_error`).

> What are your thoughts on accepting a function as a parameter to tree_weighting_method?

That seems like it's probably a good idea! Any thoughts on what would need to be passed to the weighting function?

src/sknnr/transformers/_gbnode_transformer.py
```diff
@@ -0,0 +1,271 @@
...
+    successive trees in a gradient boosting estimator.
+    """
+    # See https://github.com/lemma-osu/sknnr/issues/96#issuecomment-2967847215
+    #
+    # Calculate the initial loss, which by default in regression is based
+    # on predicting the mean for all samples. The loss function is
+    # half-squared error, so multiply by 2 to be consistent with how
+    # sklearn calculates loss.
+    initial_loss = (
+        est._loss(np.asarray(y, dtype="float64"), est._raw_predict_init(X)) * 2
```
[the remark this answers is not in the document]
aazuspan: > Will using those estimator attributes result in the same answers?

Unfortunately not.

Using `est.init_.predict` works in the default case, but if you instantiate with `init="zero"` then `_raw_predict_init` correctly returns zeros where `init_.predict` errors because `init_` is a string.

`est.loss_` is even more clear-cut since that just doesn't exist. There is a `loss` attribute, but that's always just the string name of the loss function, rather than the callable stored in `_loss`. 

I think replacing either of those private methods with public options would require duplicating a bunch of logic from `sklearn`. The other alternative would be to *only* support `oob_improvement_`, since `sklearn` already tracks that internally (including the [initial loss calculation here](https://github.com/scikit-learn/scikit-learn/blob/90338a43e8cd5e22a173ffa56e1fd93e64a08a1d/sklearn/ensemble/_gb.py#L883-L888)), but it would mean that we'd need to require subsampling to use that weighting, since otherwise there are no OOB samples.

I'm not sure what's the right approach here - I like the flexibility of being able to use the `train_score`, but the private attributes could be flaky, and maybe OOB improvement is more indicative of the actual contribution of each tree to the model?

aazuspan, line 28: I lazily multiplied the initial loss by 2 to match the default `HalfSquaredError` calculation, but I suspect we should duplicate [the logic used in `sklearn`](https://github.com/scikit-learn/scikit-learn/blob/90338a43e8cd5e22a173ffa56e1fd93e64a08a1d/sklearn/ensemble/_gb.py#L864-L873), just to make sure we don't accidentally end up accidentally doubling the initial loss relative to the other scores.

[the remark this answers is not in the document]
grovduck: For now, we will use the private attributes, as @aazuspan noted that these seem to be relatively stable in `sklearn`.  We also decided to stick with the loss associated with training data (rather than OOB) as the latter approach requires the user to specify subsampling.  Both of these decisions could be revisited at a later point.

src/sknnr/transformers/_gbnode_transformer.py
```diff
@@ -0,0 +1,271 @@
...
+    )
+    # Calculate the change in loss at each stage
+    loss_delta = np.diff(np.hstack([initial_loss, est.train_score_]))
+
+    # Normalize the loss delta to get the relative contribution of each tree
+    return (
+        np.ones_like(loss_delta, dtype="float64")
+        if np.sum(loss_delta) == 0
```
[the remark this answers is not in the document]
grovduck: Fixed in c8f3fff

grovduck:
@aazuspan, finally made a bit of progress on this PR associated with handling multi-class targets for gradient boosting classification.  I tried to write a fairly complete commit note on what had to happen, but would appreciate you checking my logic on: 1) data structures I'm using to hold the tree weights; and 2) the decision (for now) *not* to include an option to weight individual class's trees at each iteration, mainly because I'm not sure how to extract this information on loss from `sklearn`.  We still have ability to weight both the forests and the set of trees at each iteration. Here is the commit message for 416e0e3:

--------------
Gradient boosting classifiers with a multiclass target (i.e. more than two distinct labels) behave differently than either continuous regression or binary classification.  At each iteration, a separate tree is built for each class such that the final node matrix for a multi-class forest will have shape (n_samples, n_estimators, n_classes).

In order to fit the NN paradigm with Hamming distance finding, these forests must be accommodated.  This commit makes the following changes:

- Adds a new estimator attribute called `n_trees_per_iteration_` which is a list of size `self.n_forests_` and captures the number of parallel trees created per iteration.  For GB multi-class forests, this is set to n_classes (> 2).  For all other forests this is set to 1.
- Adjusts the data structure of the estimator attribute called `tree_weights_`.  Previously, this was a 2D numpy array of shape (`n_forests_`, `n_estimators`), but because multiclass forests will have (`n_estimators` * `n_classes`) trees, the weights shape may vary from forest to forest.  Now this is a list with size `n_forests_` of numpy arrays with shape (`n_estimators` * `n_trees_per_iteration_[i]`).
- Modified forest weights when applied to multi-class forests.  Because there are more trees in these forests, the forest weight needs to be adjusted (i.e. divided by `n_trees_per_iteration_[i]`) when calculating final weight.  This ensures that each forest continues to have the user-specified (or equal) weights.
- Removed `ensemble_delta_loss` and consolidated it into `GBNodeTransformer._set_tree_weights`
- Created a new test to ensure correct estimator attributes on `GBNodeTransformer` objects.
--------------
I still need to get to a few of your suggestions before it's probably ready for a full review.

aazuspan:
> Should we further say that a forest's weights should sum to 1.0? Currently, we're setting weights for uniform to be 1.0, but we could easily divide by n_estimators in the case of RF forests and by n_estimators * n_trees_per_iteration_[i] in the case of GB forests. Even though it may not be completely necessary, it might be a nice property to have?

Yeah, I like this idea! Probably not necessary as you said, but it does seem like it makes the weights easier to interpret and compare. 

> y might be an array of strings, so I think we need to first crosswalk y to its encoded values. Not fully understanding how this works, I've started with this:

Ah, good catch. It looks like `GradientBoostingClassifier` uses [`_encode_y`](https://github.com/scikit-learn/scikit-learn/blob/428a8cbdcbc0241c40209ac68653ac11b9e9e505/sklearn/ensemble/_gb.py#L1521) for the same purpose, but once `classes_` are defined your solution looks a lot simpler.

> but didn't know if that's still correct in the classification case? Do you have any insight on this? I'll admit to being a bit confused if whether the function stored in est._loss will return a comparable result in both regression and classification cases.

I think this is equivalent to how `GradientBoostingClassifier` sets its `train_score_`. I tried fitting a `GradientBoostingClassifier` with a string `y` and added a debug print where the `train_loss_` is updated [here](https://github.com/scikit-learn/scikit-learn/blob/428a8cbdcbc0241c40209ac68653ac11b9e9e505/sklearn/ensemble/_gb.py#L920), and the encoded labels passed to `loss_` match your encodings in the modified `delta_loss`. 

The only potential differences that I noticed are

1. Their encodings are floats rather than ints. That might just be a side effect of using `LabelEncoder` in `_encode_y`, and seems like it should be irrelevant.
1. They set their multiplication `factor` based on the loss function rather than hard-coding to 2.
1. They pass the optional `sample_weight` to the loss function. I haven't wrapped my head around how that might affect the loss. We could store it during `fit` and pass it on to `delta_loss`, but I'm not sure whether that's necessary or not.

grovduck:
@aazuspan, in retrospect, this was way too big of a PR.  Sorry to keep dragging you through it.  I was hoping to get a check in for the most recent changes I made, plus provide a status report of what I think is still left to do.

#### Recent changes
In the most recent [set of changes](https://github.com/lemma-osu/sknnr/compare/c46cfcb..4a95a13), I've changed the features that go into `test_estimators_with_mixed_type_forests`.  Even after the changes we made in #105 to resolve tied distances, the `gbnn-target` parametrization for this test was still failing.  Without going too deep into all the gory details, the feature that I was using for a classification target (MAX_SPECIES) was an 11-class system which, in turn, creates a separate forest for each class.  There were a couple of species that only had 1 or 2 "presence" records and it sounds like there can be some numerical instability between platforms when creating GB forests with such an unbalanced target.  So this wasn't a result of the KNeighbors Hamming distance, but actually in the construction of the individual forests (aka the transformer).

I decided the easiest tack was to shrink the number of categories to just three, holding out the top two (ABGR and TSHE) and lumping everything else into an OTHER category.  This (yet again) meant regenerating the regression files associated with that test.  Once I ran that (and was able to test on my WSL Debian instance), all neighbors and predictions matched exactly, but distances were still outside the default tolerance for `ndarrays_regression`.  The last change I made was to relax this tolerance.  

It feels a bit wrong to put so much effort into making this one test pass, so if you have other suggestions, I am all ears!

#### Roadmap
I think the roadmap going forward to completing this issue is:
- Use `sklearn` logic in calculating intial loss rather that simply multiplying by 2
- Rename `delta_loss` to `train_improvement` (per @aazuspan's suggestion) and clearly document that losses are mapped to weights and for the user to pick their loss function carefully.  For this issue, this will be the only weighting scheme we tackle, but may consider others in the future 
- Write tests to ensure that Hamming weights are standardized (sum to 1.0 across all forests and trees)
- Add documentation section for GBNN

Honestly, we probably have some additional work on getting this all to work as we would want, but I want to get it to the point where we can test whether the whole system of tree weighting is impactnull.  For that reason, I'll probably hold off on writing a section for the usage guide that goes into detail about tree/forest weights, but that would be a future PR should the weights prove useful.

aazuspan:
Thanks for the detailed update, @grovduck! I don't think either of us could have foreseen all the related issues this was going to surface, but from an observer's perspective it seems like it's gone about as smoothly as it could have, all things considered!

> It feels a bit wrong to put so much effort into making this one test pass, so if you have other suggestions, I am all ears!

Your testing changes all make sense to me. I certainly trust your approach, and I'm not worried about aggregating classes or loosening some tolerances to get things passing, especially since you have a good understanding of why those changes were needed.

Don't let me derail the PR, but just to raise this point for future discussion, I noticed that the new GBNN tests have slowed down the test suite pretty considerably: running all tests with 6 cores on my laptop went from 42.71s on `main` to 99.12s on `gbnn`. The fact that GBNN is slower than the other estimators is probably unavoidable and not a surprise, so I'm not suggesting any changes here, but we may want to think about strategies to speed up testing in general in the future. 

> I think the roadmap going forward to completing this issue is:

Your list makes sense, and I don't think I have anything to add.

> Honestly, we probably have some additional work on getting this all to work as we would want, but I want to get it to the point where we can test whether the whole system of tree weighting is impactnull. For that reason, I'll probably hold off on writing a section for the usage guide that goes into detail about tree/forest weights, but that would be a future PR should the weights prove useful.

Yeah, getting a minimal working version merged before diving too deep into documentation or fine-tuning seems like the best strategy. It would be hard to write good documentation without really understanding how the estimator works in practice.

grovduck:
@aazuspan, I think I'm ready to mark this as ready to review.  I have no illusions that this will be quick, so please take your time with all the moving parts associated with this PR.

A few follow-up notes based on the roadmap that I laid out:

#### Rename `delta_loss` to `train_improvement`
I've changed the function name and the parameter name in both `GBNNRegressor` and `GBNodeTransformer`.  I've duplicated the same "Notes" section for both of those docstrings which hopefully captures the warning you noted.

#### Use sklearn logic in calculating intial loss rather that simply multiplying by 2
I've done this as a ternary expression directly in `train_improvement`.  I initially had some failing regression tests that I couldn't quite pinpoint, but realized that in `test_estimators_with_mixed_type_forests`, we run both kinds of GB and each has a different loss function (`HalfSquaredError` for regression that has a factor of 2, `HalfMultinomialLoss` for multi-class classification that has a factor of 1).  It's a bit strange to me that a loss function with `Half` in its name would have a factor of 1, but I just followed the `sklearn` code on this one.

#### Add documentation section for GBNN
Added sections for GBNN documentation and added a fairly vague section on tree weights for now under the `RFNN and GBNN Distance Metric` section.

#### Write tests to ensure that Hamming weights are standardized (sum to 1.0 across all forests and trees)
On this, I actually had a bug in the code, where the weights array and the node arrays were the same size, but I had used `np.repeat` rather than `np.tile` to repeat weights in a multi-class problem.  Once I corrected this, the tests line up to show the relationship between forest and tree weights and that they all sum to one.  This led to new regression files being pushed (again).  Particular tests to note are:
- `test_estimators.py/test_tree_estimators_handle_forest_weights`
- `test_estimators.py/test_gbnn_multiclass_weights`
- `test_estimators.py/test_hamming_weights_sum_to_one`

#### Failing tests (still!)
The `test_estimators_with_mixed_type_forests` test is going to be the death of me.  You'll see that we still have some very slight differences in the latest CI run for this test.  I can replicate this locally between Windows and WSL and, again, it is actually different trees being built on each platform.  So rather than trying to hack a fix yet again, I'm wondering if we should xfail this particular test or continue to try to figure out how to fix it.  One thing I read was that specifying `min_samples_leaf` to be > 1 might be more forgiving on lumping multiple reference plots into the same terminal node ID which _might_ be a way to fix it.  I'd value your opinion on the path we should take here.

> Don't let me derail the PR, but just to raise this point for future discussion, I noticed that the new GBNN tests have slowed down the test suite pretty considerably ... we may want to think about strategies to speed up testing in general in the future.

I totally agree.  The multiclass classification tests are especially brutal because it has to built `n_classes` forests.  The other thing I noticed (not specifically with our test suite, but in a case where I was using lots of plots) is that by getting `independent_prediction_` and `independent_score_` during `fit` could be slowing things down as well.  It may be worth thinking about making those properties (calculated when needed) rather than estimator attributes?  But, as you say, for another PR!

(If you want to hold off review for a bit, I'll probably ask for another Copilot review just to find some stupid issues).

docs/pages/usage.md
```diff
@@ -128,11 +129,11 @@ print(y.loc[neighbor_ids[0]])
 
 ### Y-Fit Data
 
-The [GNNRegressor](api/estimators/gnn.md), [MSNRegressor](api/estimators/msn.md), and [RFNNRegressor](api/estimators/rfnn.md) estimators can be fit with `X` and `y` data, but they also accept an optional `y_fit` parameter. If provided, `y_fit` is used to fit the ordination transformer while `y` is used to fit the kNN regressor.
+The [GNNRegressor](api/estimators/gnn.md), [MSNRegressor](api/estimators/msn.md), [RFNNRegressor](api/estimators/rfnn.md), and [GBNNRegressor](api/estimators/gbnn.md) estimators can be fit with `X` and `y` data, but they also accept an optional `y_fit` parameter. If provided, `y_fit` is used to fit the ordination transformer while `y` is used to fit the kNN regressor.
```
aazuspan: Nit, but this refers to fitting the "ordination transformer", which I don't think is accurate now that the tree-based estimators are included as well. Maybe just "transformer", or "ordination/forest transformer"?

src/sknnr/transformers/_gbnode_transformer.py
```diff
@@ -0,0 +1,306 @@
...
+                weights = train_improvement(est, X, target)
+                weights /= weights.sum()
+                tree_weights.append(np.tile(weights, est.n_trees_per_iteration_))
+            return tree_weights
+
+        for est in self.estimators_:
+            arr = np.ones(
+                (est.n_estimators * est.n_trees_per_iteration_,), dtype="float64"
+            )
+            arr = arr / arr.sum(axis=0)
```
aazuspan: Minor performance nit, but could we replace this with `np.full` by pre-calculating the normalized weight?

src/sknnr/transformers/_tree_node_transformer.py
```diff
@@ -15,6 +15,15 @@
     import pandas as pd
 
 
+def uniform_weights(n_forests: int, n_estimators: int) -> list[NDArray[np.float64]]:
+    """
+    Calculate uniform weights for an ensemble of tree-based estimators.
+    """
+    return [
+        np.ones(n_estimators, dtype="float64") / n_estimators for _ in range(n_forests)
```
aazuspan: Another spot where we could potentially use `np.full` with a pre-calculated normalized weight for a minor optimization.

src/sknnr/_weighted_trees.py
```diff
@@ -69,28 +67,39 @@ def _get_hamming_weights(self):
...
+        Hamming weights should sum to 1.0 across all trees in the transformer.
         """
+        n_forests = self.transformer_.n_forests_
+
         # Equal weighting for all forests
         if isinstance(self.forest_weights, str) and self.forest_weights == "uniform":
-            forest_weights_arr = np.ones(
-                (self.transformer_.n_forests_, 1), dtype="float64"
-            )
+            forest_weights_arr = np.ones(n_forests, dtype="float64") / n_forests
```
aazuspan: Another spot where we could potentially use `np.full` with a pre-calculated normalized weight for a minor optimization.

aazuspan (changes requested):
Thanks for the detailed rundown, @grovduck! I made a first review pass, although I shied away from the Copilot comments for now. Most of my comments are the same minor performance nit in different spots, but I did find what I think is a bug in the Hamming distance weighting and algorithm selection, which I'll detail at the bottom. Other responses are below.

> I initially had some failing regression tests that I couldn't quite pinpoint, but realized that in test_estimators_with_mixed_type_forests, we run both kinds of GB and each has a different loss function (HalfSquaredError for regression that has a factor of 2, HalfMultinomialLoss for multi-class classification that has a factor of 1). It's a bit strange to me that a loss function with Half in its name would have a factor of 1, but I just followed the sklearn code on this one.

That is odd, but I agree that following `sklearn` is safe. It seems like the important thing is just that we calculate our initial loss the same way they calculate `train_score_` to avoid accidentally doubling/halving the loss delta.

> So rather than trying to hack a fix yet again, I'm wondering if we should xfail this particular test or continue to try to figure out how to fix it. One thing I read was that specifying min_samples_leaf to be > 1 might be more forgiving on lumping multiple reference plots into the same terminal node ID which might be a way to fix it. I'd value your opinion on the path we should take here.

I'm totally fine with xfailing this for now. It looks like we could even do a platform-specific xfail using a [condition](https://docs.pytest.org/en/stable/how-to/skipping.html#condition-parameter) if this is reliably passing in Windows.

`min_samples_leaf` sounds worth exploring as well, but that could be saved for a future PR. This test seems like it's been a massive headache, so I fully understand if you want to put it on the backburner for a while!

> It may be worth thinking about making those properties (calculated when needed) rather than estimator attributes? But, as you say, for another PR!

Cached properties for these sounds like a great idea!

## Weighting bug

Here's code to reproduce the bug I ran into. It looks like an incompatibility where the fit method gets set automatically to ball tree which doesn't want a `w` parameter, but that's as far as I dug.

```python
from sklearn.datasets import make_classification
from sknnr import GBNNRegressor

X, y = make_classification()
est = GBNNRegressor(n_estimators=10).fit(X, y)

# TypeError: __init__() got an unexpected keyword argument 'w'
```

The small `n_estimators` seems to be critical - I'm assuming that's used as a heuristic when selecting the fit method.

EDIT: I marked this as "request changes", but if this turns out to be a tricky bug I'm fine leaving it for a future PR.

src/sknnr/transformers/_gbnode_transformer.py
```diff
@@ -0,0 +1,306 @@
...
+    def _set_n_trees_per_iteration(self) -> list[int]:
+        return [est.n_trees_per_iteration_ for est in self.estimators_]
+
+    def _set_tree_weights(self, X, y) -> list[NDArray[np.float64]]:
+        tree_weights = []
+        if self.tree_weighting_method == "train_improvement":
+            for est, target in zip(self.estimators_, y):
+                weights = train_improvement(est, X, target)
+                weights /= weights.sum()
+                tree_weights.append(np.tile(weights, est.n_trees_per_iteration_))
```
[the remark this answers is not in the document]
grovduck: Good catch here.

It is correct that `.apply` returns nodes as shape `(n_samples, n_estimators, n_classes)` and `np.repeat` would be the correct way to align weights with scores (despite what I had written earlier).  However, `n_classes` would then be the fastest varying dimension in the tree weights rather than `n_estimators`.  My mental model is to treat each class's trees as a "forest" when apportioning forest weights (in `_get_hamming_weights`) so it's easiest to have the tree weights (of size `n_estimators`) be contiguous in the flattened weight array.  To be explicit, the flattened weights for an estimator with a regression estimator and a three-class classification estimator with `n_estimators=100` is:
```bash
[reg0_tree0 ... reg0_tree99 clf0_cls0_tree0 ... clf0_cls0_tree99 clf0_cls1_tree0 ... clf0_cls1_tree99 clf0_cls2_tree0 ... clf0_cls2_tree99]
```
This means that the `n_estimators` and `n_classes` axes in the output from `.apply` need to be swapped before reshaping to 2D (e.g. `(samples, node_ids)`) to match the weights.  Of course, this necessitated another set of new regression files.

@aazuspan, if you think there is additional information that I should include as comments in the code to help clarify this, let me know.  Part of the issue is that tree weighting and node ID capture is happening in two different places and it's a bit hard to track.

Fixed in 119f31c.

tests/test_transformers.py
```diff
@@ -292,3 +311,73 @@ def test_rfnode_transformer_non_default_parameterization(
...
+        n_samples=100,
+        n_features=20,
+        n_informative=10,
+        n_classes=n_classes,
+        random_state=42,
+    )
+
+    # Adjust y to be 2D with a multiclass target, a binary target, and a
+    # continuous target.
+    y = np.vstack([y.astype(str), y % 2, y.astype(float)], dtype="object").T
```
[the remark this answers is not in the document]
grovduck: `np.vstack` does take a `dtype` keyword argument [as of 1.2.4](https://numpy.org/doc/stable/reference/generated/numpy.vstack.html).  That seems like a pretty old requirement.  

However, since we know `y` is a 1-d array in this case, the suggested fix will work and may be more clear.  I've gone ahead and implemented that change (with unquoted `object`).

Fixed in 6be7f8c.

src/sknnr/transformers/_gbnode_transformer.py
```diff
@@ -0,0 +1,306 @@
...
+    def get_feature_names_out(self) -> NDArray:
+        check_is_fitted(self, "estimators_")
+        return np.asarray(
+            [
+                f"gb{i}_tree{j}"
+                for i in range(len(self.estimators_))
+                for j in range(self.estimators_[i].n_estimators)
+            ],
+            dtype=object,
+        )
```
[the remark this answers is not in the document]
grovduck: Another good catch.  For regression and binary classification forests, I'm following Copilot's lead of naming these `gb{i}_tree{k}` and for multi-class classification forests, I'm using `gb{i}_cls{j}_tree{k}`.  This follows the ordering as laid out in the previous comment.

Fixed in ee51728.

grovduck:
> Here's code to reproduce the bug I ran into. It looks like an incompatibility where the fit method gets set automatically to ball tree which doesn't want a `w` parameter, but that's as far as I dug.

@aazuspan, great catch.  That would have been bad to put out there!  

The good news is that I think it's a pretty simple fix and we [discussed this previously](https://github.com/lemma-osu/sknnr/issues/100#issuecomment-3667512462) when I was having all kinds of issues with Hamming weights.  I think the fix is to force `algorithm=brute` for tree-based methods and taking away that parameter (like we do with the `metric` parameter).  I don't necessarily see a downside to that given that Hamming distance doesn't play well with `kd_tree` or `ball_tree`.  Let me know if you think differently.

src/sknnr/_weighted_trees.py
```diff
@@ -69,28 +67,39 @@ def _get_hamming_weights(self):
...
-                    f"{self.transformer_.n_forests_}, but got "
-                    f"{len(self.forest_weights)}."
+                    f"{n_forests}, but got {len(self.forest_weights)}."
                 )
-            forest_weights_arr = np.asarray(
-                self.forest_weights, dtype="float64"
-            ).reshape(-1, 1)
+            forest_weights_arr = np.asarray(self.forest_weights, dtype="float64")
+            forest_weights_arr /= np.sum(forest_weights_arr)
+
```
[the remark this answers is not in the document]
grovduck: I split this out into a new `_validate_user_forest_weights` method as it was a bit more involved and didn't want to district from setting the forest weights in `_get_hamming_weights`.  I updated the `test_tree_estimator_raises_on_invalid_forest_weights` to test invalid user input.  I'm not sure my logic is as elegant as it could be (including a try/except), so let me know if you see opportunities for cleaning this up.

Fixed in 3f3f872.

docs/pages/usage.md
grovduck: Absolutely agree.  Changed "ordination transformer" to just "transformer" in two locations.

Fixed in cac1a06.

src/sknnr/transformers/_gbnode_transformer.py
grovduck: I need to get used to using `np.full`.  Fixed this and other locations in cac1a06.

grovduck:
> I'm totally fine with xfailing this for now. It looks like we could even do a platform-specific xfail using a [condition](https://docs.pytest.org/en/stable/how-to/skipping.html#condition-parameter) if this is reliably passing in Windows.

I took a bit different tack on this one and changed the test around to include different precision values for each combination.  Let me know if you think this is too funky and/or if it doesn't pass on your (linux) system.

```python
@pytest.mark.parametrize(
    ("estimator", "reference", "precision"),
    [
        # Set a higher precision threshold for GBNN with target-based forests
        # because the tree weights are less accurate when using a
        # multi-class classification forest
        (RFNNRegressor, True, 1e-8),
        (RFNNRegressor, False, 1e-8),
        (GBNNRegressor, True, 1e-8),
        (GBNNRegressor, False, 1e-2),
    ],
    ids=[
        "reference_randomForest",
        "target_randomForest",
        "reference_gbnn",
        "target_gbnn",
    ],
def test_estimators_with_mixed_type_forests(
    ndarrays_regression,
    moscow_stjoes_test_data,
    estimator,
    reference,
    precision,
):
```

src/sknnr/transformers/_tree_node_transformer.py
```diff
@@ -129,25 +139,44 @@ def _fit(self, X, y, regressor_cls, classifier_cls, reg_kwargs, clf_kwargs):
...
+        self.n_trees_per_iteration_ = self._set_n_trees_per_iteration()
+        self.tree_weights_ = self._set_tree_weights(X, y)
         return self
 
     @abstractmethod
-    def _set_tree_weights(self): ...
+    def _set_n_trees_per_iteration(self) -> list[int]: ...
+
+    @abstractmethod
+    def _set_tree_weights(self, X, y) -> list[NDArray[np.float64]]: ...
```
aazuspan, line 150: Minor naming nitpick, but from the names I would assume that these `_set_*` methods modify the attributes rather than returning. The other `_set_*` methods (e.g. `_set_dataframe_index_in` and `_set_fitted_transformer`) are mostly consistent, although now that I'm looking, `_set_estimator_types` also returns... 

You could consider renaming, but if you'd rather not expand the scope of the PR (since `_set_tree_weights` and `_set_estimator_types` are pre-existing methods), I'm fine revisiting later instead.

aazuspan:
> The good news is that I think it's a pretty simple fix and we https://github.com/lemma-osu/sknnr/issues/100#issuecomment-3667512462 when I was having all kinds of issues with Hamming weights. I think the fix is to force algorithm=brute for tree-based methods and taking away that parameter (like we do with the metric parameter). I don't necessarily see a downside to that given that Hamming distance doesn't play well with kd_tree or ball_tree. Let me know if you think differently.

It sounded familiar, but I couldn't remember where we landed - thanks for digging up the conversation. Hard-coding `algorithm` sounds like the right choice to me.

> Let me know if you think this is too funky and/or if it doesn't pass on your (linux) system.

Clever solution, and it passes on my setup!

src/sknnr/_weighted_trees.py
```diff
@@ -101,5 +96,38 @@ def _get_hamming_weights(self):
...
+            raise ValueError(
+                f"`forest_weights` must be a sequence of numeric values, "
+                f"but got {self.forest_weights} instead."
+            ) from e
+
+        if forest_weights.shape != (n_forests,):
+            raise ValueError(
+                f"Expected `forest_weights` to have length {n_forests}, "
+                f"but got {len(forest_weights)}."
+            )
```
aazuspan, line 110: I don't know how much effort you want to put into checking every possible invalid input when `forest_weights` is clearly typed as ArrayLike, but technically you can get a weird error here if you pass an int/float/bool since it creates a scalar array which has a `shape` but no length. Feel free to ignore!

```python
GBNNRegressor(forest_weights=1).fit(X, y)
# TypeError: len() of unsized object
```

src/sknnr/_weighted_trees.py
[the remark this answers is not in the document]
aazuspan: Good call splitting out the validation. It looks nice and clean, and I think the try/except is appropriate since we expect it to pass in normal circumstances.

I did notice there's no positive sum check in the validation, so it is still possible to end up with NaN weights if a user does something stupid like `forest_weights=[0, 0, 0]`.

src/sknnr/transformers/_gbnode_transformer.py
```diff
@@ -0,0 +1,310 @@
...
+    def _set_tree_weights(self, X, y) -> list[NDArray[np.float64]]:
+        tree_weights = []
+        if self.tree_weighting_method == "train_improvement":
+            for est, target in zip(self.estimators_, y):
+                weights = train_improvement(est, X, target)
+                weights /= weights.sum()
+                tree_weights.append(np.tile(weights, est.n_trees_per_iteration_))
+            return tree_weights
+
+        for est in self.estimators_:
```
aazuspan: I was a little confused when I ran some old code that still used `tree_weighting_method="delta_loss"` and it didn't break, but it looks like this implicitly treats any unrecognized methods as uniform? It might be worth making that check explicit and failing on unrecognized options, just in case.

src/sknnr/transformers/_gbnode_transformer.py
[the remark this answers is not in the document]
aazuspan: The comment you added is very helpful! I always get a little lost trying to track dimensions of the weights and node IDs, but I'm not sure how much more you could do, since it seems like an unavoidably complex problem, especially in the multi-class case.

grovduck:
@aazuspan, I've taken out `algorithm` (and `leaf_size` which was a `BallTree`/`KDTree` specific parameter) from both tree-based estimators and the `WeightedTreesNNRegressor` superclass.  I _think_ this is ready for another review when you get the chance.  Thanks!

aazuspan (changes requested):
This is looking great, @grovduck! I just had a few more comments, but all pretty minor.

src/sknnr/transformers/_tree_node_transformer.py
grovduck: Yep, good catch.  This didn't even occur to me, but I agree.  I've created a new issue (#107) to track this rather than bury the fix in this PR.

src/sknnr/_weighted_trees.py
grovduck: You're absolutely right.  I've changed the error message to report `{forest_weights.size}` since it's guaranteed to be a numpy array at that point.  Fixed in 770c841 with accompanying tests.

src/sknnr/_weighted_trees.py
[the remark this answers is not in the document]
grovduck: D'oh, I thought my test for non-negative was sufficient, but obviously forgot about this case!  Fixed in 770c841 with accompanying test.

src/sknnr/transformers/_gbnode_transformer.py
grovduck: Agreed.  Better to be safe.  Fixed in 770c841.

src/sknnr/transformers/_tree_node_transformer.py
aazuspan: Sounds good!

aazuspan (approved):
Ready to merge whenever you are! This was a huge lift with all the tricky edge cases and reproducibility/performance hurdles, and it turned out great. I'm excited to start testing out GBNN for real 🎉

grovduck:
Whoo, this one was an absolute beast (started in August!).  Thank you so much for persevering through all the pain.  I'm not sure I'm ready to call it "great" 😄, but it's a solid start.  Onward!

PR #103, opened 2026-01-09
Remove invalid `undoc_members` config option

### LINKED ISSUE #102: Fix RTD build failures caused by invalid configuration option
ReadTheDocs builds [are failing](https://app.readthedocs.org/projects/sknnr/builds/30950283/) due to an unsupported configuration option:

> ERROR   -  Invalid options: PythonOptions.__init__() got an unexpected keyword argument 'undoc_members'

This came up before (possibly in `sklearn-raster`, although I didn't find it from a quick search), and `undoc_members` is a [Sphinx configuration option](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-option-automodule-undoc-members) that was accidentally added to `mkdocs.yml` and needs to be removed. I'm not sure why it's only now causing build failures, but it's an easy fix.

aazuspan:
Closes #102 by removing the invalid `undoc_members` configuration option.

`undoc_members` seems to be a Sphinx configuration option that was accidentally added. The parameter was previously being ignored, but now causes build errors locally and [in RTD](https://app.readthedocs.org/projects/sknnr/builds/30950283/).

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -57,7 +57,6 @@ plugins:
           options:
             show_source: false
             inherited_members: true
-            undoc_members: true
             docstring_style: numpy
             show_if_no_docstring: true
             show_signature_annotations: true


grovduck (approved):
LGTM!  Thanks for jumping on this!

PR #108, opened 2026-03-20
Upgrade minimum Python to 3.11

aazuspan:
This upgrades the minimum Python from 3.9 to 3.11 following the [NEP 29 schedule](https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule), which also resolves a [Hatch bug in 3.9](https://github.com/lemma-osu/sklearn-raster/pull/109#issuecomment-3987828351). Local and Github Actions now test through 3.14. Pre-commit hook and action versions are bumped to the latest.

Aside from updating project and CI metadata, there were a few typing changes as well as the new `zip(..., strict=True)` parameter to satisfy static analysis.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,10 +14,10 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
 
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v6
         with:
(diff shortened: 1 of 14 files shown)


aazuspan:
I'll go ahead and merge since we already discussed the version change and there weren't any significant side effects. The only functional changes are the strict zips that would now error rather than silently truncating, but I'm confident that was the original intent throughout.

grovduck (commented):
@aazuspan, this all looks good to me.  I can confirm that the use of `zip(..., strict=True)` is appropriate for all uses.  Thx.

PR #112, opened 2026-03-23
Expand typing for estimators

aazuspan:
Along with #109 and #110, this is another step (maybe the final step?) towards #7, by adding typing for all estimators. The only modules I haven't currently touched are the `_cca` and `_ccora` implementations, since I know there have been some significant efforts to refactor those in #60 that might influence how we approach typing.

This PR includes more changes and decisions that aren't strictly typing than the other two, so I'll outline those below.

## Mixin class inheritance

Correctly typing the mixins was tricky, since they rely on methods and attributes of the class that they expect to be mixed into. I played around with typing `self` as a protocol, but found that required repeating a lot of our base classes as protocols, didn't work for every case, and only enforced typing when methods are called.

Instead, I ended up loosely following the recommendations in [this blog post](https://adamj.eu/tech/2025/05/01/python-type-hints-mixin-classes/), which suggests using shared base classes to satisfy type checks. 

For example, `IndependentPredictorMixin` now inherits from `KNeighborsRegressor`, which ensures that it implements `predict` and `score` with optional `X` arguments. `YFitMixin` inherits from `TransformedKNeighborsRegressor`, since that defines the `_get_transformer` method (that also required changing the definition order). I don't think this leads to any functional changes, but it's worth thinking about MRO a little more to make sure it doesn't break anything.

## DataLike

Another challenge was coming up with a good type to represent the X and y inputs for estimator methods. For now, I'm using a union of Numpy `ArrayLike` and the `DataFrameLike` and `SeriesLike` protocols that I called `DataLike`, but I'm not totally sure that's the right definition or the right name. 

The [Microsoft typestubs for scikit-learn](https://github.com/microsoft/python-type-stubs/blob/main/stubs/sklearn/_typing.pyi) seem to use `MatrixLike` for `X` and `MatrixLike | ArrayLike` for `y`, where `MatrixLike` is a union of Numpy `ArrayLike`, `pd.DataFrame`, and `scipy.sparse.spmatrix`. Maybe we should go with something like that, but replacing the specific Pandas type with the `DataFrameLike` protocol? I also wasn't sure whether we should name our type `MatrixLike` for consistency, or change it to avoid confusion with our slightly different definition.

## Type-checking rules

This PR involved a lot of new imports, most of which are only used for typing. To help stay consistent and avoid typing-only imports at runtime, I added the [flake8-type-checking](https://docs.astral.sh/ruff/rules/#flake8-type-checking-tc) rules to Ruff.

## PR order

I made this PR on top of #110 since it builds on the `types` module introduced there. I'll leave this is a draft until that's merged, then rebase this as needed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -92,6 +92,7 @@ select = [
     "RET",
     "SIM",
     "PERF",
+    "TC",
 ]
 
 [tool.ruff.lint.isort]
--- a/src/sknnr/_base.py
+++ b/src/sknnr/_base.py
@@ -1,21 +1,21 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from collections.abc import Callable
 from typing import TYPE_CHECKING
 
 import numpy as np
 from sklearn.base import BaseEstimator, TransformerMixin
-from sklearn.metrics import DistanceMetric
 from sklearn.neighbors import KNeighborsRegressor
 from sklearn.utils.validation import _is_arraylike, check_is_fitted
 
 from sknnr.utils import is_dataframe_like, is_series_like
 
 if TYPE_CHECKING:
+    from collections.abc import Callable
     from typing import Literal, Self
 
     from numpy.typing import NDArray
+    from sklearn.metrics import DistanceMetric
     from sklearn.utils._tags import Tags
 
     from .transformers._base import ComponentReducerMixin
--- a/src/sknnr/_base.py
+++ b/src/sknnr/_base.py
@@ -174,7 +174,7 @@ class RawKNNRegressor(
         return_distance: bool = True,
         return_dataframe_index: bool = False,
         use_deterministic_ordering: bool = True,
-    ) -> NDArray | tuple[NDArray, NDArray]:
+    ) -> NDArray[np.int64] | tuple[NDArray[np.float64], NDArray[np.int64]]:
         """
         Find the K-neighbors of a point or points in the dataset and optionally
         return dataframe indexes rather than array indices when the model was
--- a/src/sknnr/_base.py
+++ b/src/sknnr/_base.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from collections.abc import Callable
-from typing import TYPE_CHECKING, Literal
+from typing import TYPE_CHECKING
 
 import numpy as np
 from sklearn.base import BaseEstimator, TransformerMixin
@@ -11,10 +11,19 @@ from sklearn.neighbors import KNeighborsRegressor
 from sklearn.utils.validation import _is_arraylike, check_is_fitted
 
 if TYPE_CHECKING:
+    from typing import Literal
+
+    from numpy.typing import NDArray
+
     from .transformers._base import ComponentReducerMixin
 
 
-def _validate_data(estimator, *, ensure_all_finite: bool = True, **kwargs):
+def _validate_data(
+    estimator: BaseEstimator,
+    *,
+    ensure_all_finite: bool | Literal["allow-nan"] = True,
+    **kwargs,
+) -> NDArray | tuple[NDArray, NDArray]:
     """
     Compatibility wrapper around sklearn's _validate_data function.
 
--- a/src/sknnr/_base.py
+++ b/src/sknnr/_base.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, overload
 
 import numpy as np
 from sklearn.base import BaseEstimator, TransformerMixin
@@ -20,9 +20,40 @@ if TYPE_CHECKING:
     from .types import DataLike
 
 
+@overload
 def _validate_data(
     estimator: BaseEstimator,
     *,
+    X: DataLike,
+    y: Literal["no-validation"] = "no-validation",
+    ensure_all_finite: bool | Literal["allow-nan"] = True,
+    **kwargs,
+) -> NDArray: ...
+@overload
+def _validate_data(
+    estimator: BaseEstimator,
+    *,
+    X: Literal["no-validation"] = "no-validation",
+    y: DataLike,
+    ensure_all_finite: bool | Literal["allow-nan"] = True,
+    **kwargs,
+) -> NDArray: ...
+@overload
+def _validate_data(
+    estimator: BaseEstimator,
+    *,
+    X: DataLike,
+    y: DataLike,
+    ensure_all_finite: bool | Literal["allow-nan"] = True,
+    **kwargs,
+) -> tuple[NDArray, NDArray]: ...
+
+
+def _validate_data(
+    estimator: BaseEstimator,
+    *,
+    X: DataLike | Literal["no-validation"] = "no-validation",
+    y: DataLike | Literal["no-validation"] | None = "no-validation",
     ensure_all_finite: bool | Literal["allow-nan"] = True,
     **kwargs,
 ) -> NDArray | tuple[NDArray, NDArray]:
@@ -39,9 +70,13 @@ def _validate_data(
     try:
         from sklearn.utils.validation import validate_data
     except ImportError:
-        return estimator._validate_data(force_all_finite=ensure_all_finite, **kwargs)
+        return estimator._validate_data(
+            X=X, y=y, force_all_finite=ensure_all_finite, **kwargs
+        )
 
-    return validate_data(estimator, ensure_all_finite=ensure_all_finite, **kwargs)
+    return validate_data(
+        estimator, X=X, y=y, ensure_all_finite=ensure_all_finite, **kwargs
+    )
 
 
 class DFIndexCrosswalkMixin:
--- a/src/sknnr/_base.py
+++ b/src/sknnr/_base.py
@@ -348,7 +348,7 @@ class TransformedKNeighborsRegressor(BaseEstimator, ABC):
         return_distance: bool = True,
         return_dataframe_index: bool = False,
         use_deterministic_ordering: bool = True,
-    ) -> NDArray | tuple[NDArray, NDArray]:
+    ) -> NDArray[np.int64] | tuple[NDArray[np.float64], NDArray[np.int64]]:
         """
         Find the K-neighbors of a point or points of transformed feature data
         and optionally return dataframe indexes rather than array indices when
@@ -402,7 +402,7 @@ class TransformedKNeighborsRegressor(BaseEstimator, ABC):
             use_deterministic_ordering=use_deterministic_ordering,
         )
 
-    def predict(self, X: DataLike) -> NDArray:
+    def predict(self, X: DataLike) -> NDArray[np.float64]:
         X_transformed = self._transform_X(X)
         return self.regressor_.predict(X_transformed)
 
--- a/src/sknnr/_base.py
+++ b/src/sknnr/_base.py
@@ -8,8 +8,6 @@ from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.neighbors import KNeighborsRegressor
 from sklearn.utils.validation import _is_arraylike, check_is_fitted
 
-from sknnr.utils import is_dataframe_like, is_series_like
-
 if TYPE_CHECKING:
     from collections.abc import Callable
     from typing import Literal, Self
@@ -51,10 +49,9 @@ class DFIndexCrosswalkMixin:
 
     def _set_dataframe_index_in(self, X: DataLike) -> None:
         """Store dataframe indexes if X is a dataframe."""
-        if is_dataframe_like(X) or is_series_like(X):
-            index = X.index
-            if _is_arraylike(index):
-                self.dataframe_index_in_ = np.asarray(index)
+        index = getattr(X, "index", None)
+        if _is_arraylike(index):
+            self.dataframe_index_in_ = np.asarray(index)
 
 
 class IndependentPredictorMixin(KNeighborsRegressor):
--- a/src/sknnr/_base.py
+++ b/src/sknnr/_base.py
@@ -374,7 +374,7 @@ class TransformedKNeighborsRegressor(BaseEstimator, ABC):
         X_transformed = self._transform_X(X)
         return self.regressor_.predict(X_transformed)
 
-    def score(self, X: DataLike, y: DataLike | None = None) -> float:
+    def score(self, X: DataLike, y: DataLike) -> float:
         X_transformed = self._transform_X(X)
         return self.regressor_.score(X_transformed, y)
 
--- a/src/sknnr/_base.py
+++ b/src/sknnr/_base.py
@@ -10,12 +10,16 @@ from sklearn.metrics import DistanceMetric
 from sklearn.neighbors import KNeighborsRegressor
 from sklearn.utils.validation import _is_arraylike, check_is_fitted
 
+from sknnr.utils import is_dataframe_like, is_series_like
+
 if TYPE_CHECKING:
-    from typing import Literal
+    from typing import Literal, Self
 
     from numpy.typing import NDArray
+    from sklearn.utils._tags import Tags
 
     from .transformers._base import ComponentReducerMixin
+    from .types import DataLike
 
 
 def _validate_data(
@@ -45,37 +49,22 @@ def _validate_data(
 class DFIndexCrosswalkMixin:
     """Mixin to crosswalk array indices to dataframe indexes."""
 
-    def _set_dataframe_index_in(self, X):
+    def _set_dataframe_index_in(self, X: DataLike) -> None:
         """Store dataframe indexes if X is a dataframe."""
-        index = getattr(X, "index", None)
-        if _is_arraylike(index):
-            self.dataframe_index_in_ = np.asarray(index)
+        if is_dataframe_like(X) or is_series_like(X):
+            index = X.index
+            if _is_arraylike(index):
+                self.dataframe_index_in_ = np.asarray(index)
 
 
-class IndependentPredictorMixin:
+class IndependentPredictorMixin(KNeighborsRegressor):
     """Mixin to return independent predictions based on the X data used
     for fitting the model."""
 
-    def _set_independent_prediction_attributes(self, y):
+    def _set_independent_prediction_attributes(self, y: DataLike) -> None:
         """Store independent predictions and score."""
         self.independent_prediction_ = super().predict(X=None)
         self.independent_score_ = super().score(X=None, y=y)
-
-
-class YFitMixin:
-    """Mixin for transformed estimators that use an optional y_fit to fit their
-    transformer."""
-
-    def _set_fitted_transformer(self, X, y):
-        """Fit and store the transformer, using stored y_fit data if available."""
-        y_fit = self.y_fit_ if self.y_fit_ is not None else y
-        self.transformer_ = self._get_transformer().fit(X, y_fit)
-
-    def fit(self, X, y, y_fit=None):
-        """Fit using transformed feature data. If y_fit is provided, it will be used
-        to fit the transformer."""
-        self.y_fit_ = y_fit
-        return super().fit(X, y)
 
 
 class RawKNNRegressor(
@@ -139,7 +128,7 @@ class RawKNNRegressor(
 
     DISTANCE_PRECISION_DECIMALS = 10
 
-    def fit(self, X, y):
+    def fit(self, X: DataLike, y: DataLike) -> Self:
         """Override fit to set attributes using mixins."""
         self._set_dataframe_index_in(X)
         self = super().fit(X, y)
@@ -148,12 +137,12 @@ class RawKNNRegressor(
 
     def kneighbors(
         self,
-        X=None,
-        n_neighbors=None,
-        return_distance=True,
-        return_dataframe_index=False,
-        use_deterministic_ordering=True,
-    ):
+        X: DataLike | None = None,
+        n_neighbors: int | None = None,
+        return_distance: bool = True,
+        return_dataframe_index: bool = False,
+        use_deterministic_ordering: bool = True,
+    ) -> NDArray | tuple[NDArray, NDArray]:
         """
         Find the K-neighbors of a point or points in the dataset and optionally
         return dataframe indexes rather than array indices when the model was
@@ -261,22 +250,22 @@ class TransformedKNeighborsRegressor(BaseEstimator, ABC):
         subclasses."""
         ...
 
-    def _set_fitted_transformer(self, X, y):
+    def _set_fitted_transformer(self, X: DataLike, y: DataLike) -> None:
         """Fit and store the transformer."""
         self.transformer_ = self._get_transformer().fit(X, y)
 
-    def _get_additional_regressor_init_kwargs(self) -> dict:
+    def _get_additional_regressor_init_kwargs(self) -> dict[str, object]:
         """Get any additional keyword arguments for the KNeighbors regressor
         initialization. Subclasses can override to provide additional arguments.
         """
         return {}
 
-    def _transform_X(self, X):
+    def _transform_X(self, X: DataLike | None) -> DataLike | None:
         """Transform feature data using the fitted transformer."""
         check_is_fitted(self, "transformer_")
         return self.transformer_.transform(X) if X is not None else X
 
-    def fit(self, X, y):
+    def fit(self, X: DataLike, y: DataLike) -> Self:
         """Fit using transformed feature data."""
         _validate_data(self, X=X, y=y, ensure_all_finite=True, multi_output=True)
 
@@ -322,12 +311,12 @@ class TransformedKNeighborsRegressor(BaseEstimator, ABC):
 
     def kneighbors(
         self,
-        X=None,
-        n_neighbors=None,
-        return_distance=True,
-        return_dataframe_index=False,
-        use_deterministic_ordering=True,
-    ):
+        X: DataLike | None = None,
+        n_neighbors: int | None = None,
+        return_distance: bool = True,
+        return_dataframe_index: bool = False,
+        use_deterministic_ordering: bool = True,
+    ) -> NDArray | tuple[NDArray, NDArray]:
         """
         Find the K-neighbors of a point or points of transformed feature data
         and optionally return dataframe indexes rather than array indices when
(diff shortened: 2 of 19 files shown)


src/sknnr/types.py
```diff
@@ -11,15 +13,31 @@
...
+
+
+DataLike = DataFrameLike | SeriesLike | ArrayLike
+"""
+Data structures that can be used as input to Scikit-learn estimators, including labeled
+and unlabeled arrays.
+
+This broadly follows the "array-like" definition from Scikit-learn:
+https://scikit-learn.org/stable/glossary.html#term-array-like
+"""
```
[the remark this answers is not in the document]
aazuspan: As far as I know, this is the accepted (if not official) way to document module-level variables, and it is correctly parsed by my LSP as a docstring for `DataLike` and appears on hover.

Bugbear doesn't have a problem with it.

src/sknnr/transformers/_gbnode_transformer.py
```diff
@@ -1,16 +1,23 @@
...
 from sklearn.ensemble import GradientBoostingClassifier, GradientBoostingRegressor
 from sklearn.utils.validation import check_is_fitted
 
 from ._tree_node_transformer import TreeNodeTransformer
 
+if TYPE_CHECKING:
+    from typing import Literal, Self
+
+    from numpy.typing import NDArray
+    from sklearn.base import BaseEstimator
```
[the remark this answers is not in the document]
aazuspan: Fixed. I'm not sure why it wasn't flagged.

pyproject.toml
```diff
@@ -88,6 +88,7 @@ select = [
     "RET",
     "SIM",
     "PERF",
+    "TC",
```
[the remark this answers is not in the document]
aazuspan: Adding TC was obviously intentional. I fixed the runtime import of `NDArray` in `_cca`. I suspect that wasn't being flagged because the module didn't have a `from __future__ import annotations` import, which did technically make it a runtime dependency.

src/sknnr/transformers/_rfnode_transformer.py
```diff
@@ -1,16 +1,23 @@
...
 
 from ._tree_node_transformer import TreeNodeTransformer, uniform_weights
 
+if TYPE_CHECKING:
+    from collections.abc import Callable
+    from typing import Literal, Self
+
+    from numpy.random import RandomState
+    from numpy.typing import NDArray
+
```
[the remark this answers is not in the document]
aazuspan: Fixed

src/sknnr/_base.py
```diff
@@ -36,39 +49,24 @@ def _validate_data(estimator, *, ensure_all_finite: bool = True, **kwargs):
...
     """Mixin to crosswalk array indices to dataframe indexes."""
 
-    def _set_dataframe_index_in(self, X):
+    def _set_dataframe_index_in(self, X: DataLike) -> None:
         """Store dataframe indexes if X is a dataframe."""
-        index = getattr(X, "index", None)
-        if _is_arraylike(index):
-            self.dataframe_index_in_ = np.asarray(index)
+        if is_dataframe_like(X) or is_series_like(X):
+            index = X.index
```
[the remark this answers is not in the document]
aazuspan: Good point. I reverted this change.

src/sknnr/utils/__init__.py
```diff
@@ -4,13 +4,20 @@
...
 
     import pandas as pd
 
-    from ..types import DataFrameLike, DTypeLike, SeriesLike
+
+def _has_all_attrs(obj: object, protocol: type) -> bool:
+    """
+    Check if an object satisfies the attributes of a protocol.
+    """
+    return all(hasattr(obj, attr) for attr in protocol.__annotations__)
```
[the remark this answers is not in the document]
aazuspan: Good point - I reverted to the original checks. You can make a protocol [run-time checkable with isinstance](https://typing.python.org/en/latest/spec/protocol.html#runtime-checkable-decorator-and-narrowing-types-by-isinstance), but for now I went with the simple solution.

src/sknnr/types.py
```diff
@@ -11,15 +13,31 @@
...
     columns: Sequence[Hashable]
     dtypes: Sequence[DTypeLike]
 
 
-class SeriesLike(Protocol):
+class SeriesLike(Indexed, Protocol):
     """A protocol for series-like objects, such as pandas and polars Series."""
 
     name: Hashable | None
     dtype: DTypeLike
```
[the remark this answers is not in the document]
aazuspan: Fixed by reverting the change

src/sknnr/transformers/_tree_node_transformer.py
```diff
@@ -61,27 +72,27 @@ def _validate_and_promote_targets(
...
+                    }
+                ):
+                    raise ValueError(
+                        f"Target {name} has non-string types ({non_string_types}) "
+                        f"that cannot be safely converted to a string dtype "
+                        f"({promoted_dtype})."
+                    )
+
+                # Otherwise, promote the target to the minimum numpy dtype
                 target = target.astype(promoted_dtype)
```
aazuspan, line 96: I had to wrap this block in a type guard to convince mypy that it's not a categorical type. That feels redundant since we already checked the categorical case, but my understanding is that type guards only narrow in the positive case (i.e. mypy knows it's a categorical in the `if` block, but doesn't know it's *not* a categorical in the `else` block). There's a more powerful [TypeIs](https://peps.python.org/pep-0742/) that would solve this, but it's not available until 3.13. 

I *think* this is effectively equivalent to the previous behavior at runtime since we've already restricted to just Numpy types and categorical when we got the promoted types, but it's possible I missed a nuance. Let me know if this seems problematic, or if I should add a comment here.

src/sknnr/_base.py
```diff
@@ -372,21 +367,37 @@ def kneighbors(
...
 
-    def __sklearn_tags__(self):
+    def __sklearn_tags__(self) -> Tags:
         tags = super().__sklearn_tags__()
         tags.input_tags.sparse = False
 
         return tags
 
 
+class YFitMixin(TransformedKNeighborsRegressor):
```
aazuspan, line 420: I had to move this definition now that it inherits from `TransformedKNeighborsRegressor`.

src/sknnr/types.py
```diff
@@ -1,25 +1,44 @@
...
 
-    from numpy.typing import DTypeLike as NDTypeLike
+    from pandas import CategoricalDtype
     from pandas.api.extensions import ExtensionDtype
 
-DTypeLike = Union["NDTypeLike", "ExtensionDtype"]
+NumpyDTypeLike: TypeAlias = DTypeLike
+"""Any data type that can be used in a numpy array."""
+
+AnyDTypeLike = Union[NumpyDTypeLike, "ExtensionDtype"]
```
aazuspan, line 22: I renamed from `DTypeLike` to `AnyDTypeLike` to avoid confusion with the unaliased `numpy.typing.DTypeLike`, since my LSP otherwise didn't distinguish between the two.

src/sknnr/types.py
```diff
@@ -1,25 +1,44 @@
...
+"""Any Numpy or Pandas data type, including extension types."""
+
+DataDTypeLike = Union[NumpyDTypeLike, "CategoricalDtype"]
+"""Data types used for features or targets, i.e. Numpy and Pandas categorical dtypes."""
```
aazuspan, line 26: I named this to match `DataLike`. If we find a better name for that, we should rename this too.

Also, is it an oversimplification to say that features/targets must be either Numpy types or categorical? I think that's effectively how it works in `_TreeNodeTransformer`, but it might not be strictly accurate. Should we make this more specific, like `NumpyOrCategoricalDType`?

aazuspan (commented):
@grovduck, I think this is at the point where it's reviewable now - I've rebased from `main` and confirmed that it's not adding any new typing errors once we include the mypy dependencies.

This one is a lot less targeted than the other PRs and has more runtime implications, so I suspect it'll take a while for us to work through. I tried to point out the spots I'm particularly worried about in the original post or in review comments, but please keep an eye out in case I broke anything else, or mis-represented any types (in particular, I struggled to keep track of where `X` and `y` parameters were or weren't optional).

src/sknnr/transformers/_cca.py
grovduck, line 1: This makes sense why you had to add this with the new `TC` ruff check.  At some point, I'll get back to this beast!

src/sknnr/types.py
grovduck: `AnyDTypeLike` seems like an appropriate name here.

src/sknnr/types.py
grovduck: > I named this to match `DataLike`. If we find a better name for that, we should rename this too.

Right, `DataLike` and `DataDTypeLike` are fairly generic, but that doesn't really bother me in this situation.  Your new docstrings help clarify and expand upon the intent, which is available to the user when hovering over.

> Also, is it an oversimplification to say that features/targets must be either Numpy types or categorical? I think that's effectively how it works in _TreeNodeTransformer, but it might not be strictly accurate.

AFAIK, this is accurate.  In looking at the [scikit-learn page](https://scikit-learn.org/stable/glossary.html#term-array-like) you linked, it seems like you're just expanding their `array-like` type to explicitly include `CategoricalDtype`.  I think you're correct that these are the types that are supported in `TreeNodeTransformer`.

src/sknnr/transformers/_base.py
```diff
@@ -37,11 +48,11 @@ class StandardScalerWithDOF(StandardScaler):
         The number of samples processed by the estimator for each feature.
     """
 
-    def __init__(self, ddof=0):
+    def __init__(self, ddof: int = 0):
```
grovduck, line 51: Just a quick verification, but `__init__` doesn't necessarily have to show a return value of `None`.  Is that implicit?

src/sknnr/transformers/_tree_node_transformer.py
```diff
@@ -109,7 +120,15 @@ def _set_estimator_types(
...
+    def _fit(
+        self,
+        X: DataLike,
+        y: DataLike,
+        regressor_cls,
+        classifier_cls,
+        reg_kwargs,
+        clf_kwargs,
+    ) -> Self:
         X = _validate_data(self, X=X, reset=True)
```
grovduck: This isn't in the scope of changes, but I came across it when looking at `_set_tree_weights`.  You've typed the return value of `_validate_data` to be `NDArray | tuple[NDArray, NDArray]` which makes sense because either `X` or `Y` or both is returned.  But then `X` has that type and when used in `_set_tree_weights`, it's expecting that `X` is `NDArray`.

I don't know if this is an issue or not, given that the argument can be either of the two types (`NDArray | tuple[NDArray, NDArray]`)  but the parameter is only `NDArray`.  Is this a situation where you would explicitly type the variable?
```python
X: NDArray = _validate_data(self, X=X, reset=True)
```
(I also noticed that we have a `TODO` on `_validate_data` to rework it when `scikit-learn >= 1.6.0`.  I can't remember when we did that, but that's what shows in our `pyproject.toml' file.)

src/sknnr/utils/__init__.py
```diff
@@ -52,9 +60,26 @@ def is_nan_like(x: object) -> bool:
...
+        return True
+    except TypeError:
+        return False
+
+
+def is_categorical_dtype(t: AnyDTypeLike) -> TypeGuard[CategoricalDtype]:
+    """
+    Check if `t` is a categorical dtype using duck-typing.
+    """
+    return str(t) == "category"
```
grovduck, line 79: Excellent, these are both nice to have!

src/sknnr/utils/__init__.py
```diff
@@ -29,18 +38,17 @@ def is_series_like(obj: object) -> TypeGuard[SeriesLike]:
     return hasattr(obj, "name") and hasattr(obj, "dtype")
 
 
-def is_number_like_type(t: DTypeLike) -> bool:
+def is_number_like_type(t: AnyDTypeLike) -> bool:
```
grovduck: Should this be renamed to `is_number_like_dtype`?

src/sknnr/transformers/_tree_node_transformer.py
```diff
@@ -61,27 +72,27 @@ def _validate_and_promote_targets(
             # minimum numpy dtype.  Numpy does not support categorical dtypes,
             # but we need to retain the categorical dtype label to correctly route
             # the target to a random forest classifier.
```
grovduck: Can we generalize the comment here to be "the target to a tree-based classifier".

src/sknnr/transformers/_tree_node_transformer.py
grovduck: > I had to wrap this block in a type guard to convince mypy that it's not a categorical type. That feels redundant since we already checked the categorical case, but my understanding is that type guards only narrow in the positive case (i.e. mypy knows it's a categorical in the `if` block, but doesn't know it's _not_ a categorical in the `else` block)

That makes sense.  I had to read up on `TypeGuard`, but I understand it in the same way you do.

> I think this is effectively equivalent to the previous behavior at runtime since we've already restricted to just Numpy types and categorical when we got the promoted types, but it's possible I missed a nuance. Let me know if this seems problematic, or if I should add a comment here.

I think you're fine here, unless I'm missing the same nuance.

src/sknnr/transformers/_tree_node_transformer.py
```diff
@@ -109,7 +120,15 @@ def _set_estimator_types(
...
 
-    def _fit(self, X, y, regressor_cls, classifier_cls, reg_kwargs, clf_kwargs):
+    def _fit(
+        self,
+        X: DataLike,
+        y: DataLike,
+        regressor_cls,
+        classifier_cls,
+        reg_kwargs,
+        clf_kwargs,
```
grovduck: Should these parameters have types?

src/sknnr/_base.py
```diff
@@ -139,12 +134,12 @@ def fit(self, X, y):
...
-        return_distance=True,
-        return_dataframe_index=False,
-        use_deterministic_ordering=True,
-    ):
+        X: DataLike | None = None,
+        n_neighbors: int | None = None,
+        return_distance: bool = True,
+        return_dataframe_index: bool = False,
+        use_deterministic_ordering: bool = True,
+    ) -> NDArray | tuple[NDArray, NDArray]:
```
grovduck: I've noticed throughout this PR that you're generally just using `NDArray` without more specific typing (like `NDArray[np.int64]`).  I know that some of the (rare) typing that I did specialized the element type as well.  If you'd rather just leave as `NDArray`, I'm OK with that.

src/sknnr/_base.py
grovduck: Yep, that makes sense.

grovduck (commented):
@aazuspan, wow, this was a fair bit of work!  Thank you for tackling it.  I'll admit to some glassy eyes by the end of the review, because similar changes were made in a lot of places, so hopefully I caught any issues.

> For example, `IndependentPredictorMixin` now inherits from `KNeighborsRegressor`, which ensures that it implements `predict` and `score` with optional `X` arguments. `YFitMixin` inherits from `TransformedKNeighborsRegressor`, since that defines the `_get_transformer` method (that also required changing the definition order). I don't think this leads to any functional changes, but it's worth thinking about MRO a little more to make sure it doesn't break anything.

That was a cool blog post and made good sense.  These look like good changes to me.

> The [Microsoft typestubs for scikit-learn](https://github.com/microsoft/python-type-stubs/blob/main/stubs/sklearn/_typing.pyi) seem to use `MatrixLike` for `X` and `MatrixLike | ArrayLike` for `y`, where `MatrixLike` is a union of Numpy `ArrayLike`, `pd.DataFrame`, and `scipy.sparse.spmatrix`. Maybe we should go with something like that, but replacing the specific Pandas type with the `DataFrameLike` protocol? I also wasn't sure whether we should name our type `MatrixLike` for consistency, or change it to avoid confusion with our slightly different definition.

The one thing that gives me pause here is the support for sparse matrices in `MatrixLike`?  For `TransformedKNeighborsRegressor`, we explicitly have `tags.input_tags.sparse = False`.  Maybe I'm confused on whether that would cause an issue?  If we typed `X` as `MatrixLike` and provided a `scipy.sparse.spmatrix`, then wouldn't any of our estimators derived from `TransformedKNeighborsRegressor` fail to fit?

> To help stay consistent and avoid typing-only imports at runtime, I added the [flake8-type-checking](https://docs.astral.sh/ruff/rules/#flake8-type-checking-tc) rules to Ruff.

I like this!  And potentially can speed-up runtimes?  I think I've read that from you.

> This one is a lot less targeted than the other PRs and has more runtime implications, so I suspect it'll take a while for us to work through. I tried to point out the spots I'm particularly worried about in the original post or in review comments, but please keep an eye out in case I broke anything else, or mis-represented any types (in particular, I struggled to keep track of where X and y parameters were or weren't optional).

I actually thought the PR was really clean, unless I'm missing something, which you shouldn't put past me.  I did notice the `y: None = None` typing on all the `X`-only transformers, so that seems solid to me!

I have a few comments in here that probably need to be addressed, so I'll leave as just a comment for now, but happy to take a second look when you're ready.

src/sknnr/transformers/_base.py
aazuspan: Good question - I'd never really looked into this before. The [conclusion to this long mypy thread](https://github.com/python/mypy/issues/604#issuecomment-348525995) seems to be that it depends on whether `__init__` has any typed arguments. If so, `-> None` is inferred and the constructor is type-checked. If not, you do need an explicit `-> None` to enable type-checking.

Since all of our constructors take arguments, the return type *is* currently implicit. I'm happy to go explicit if you have a preference.

src/sknnr/transformers/_tree_node_transformer.py
aazuspan: Done!

src/sknnr/utils/__init__.py
aazuspan: Yeah, I changed that and the docstring for consistency with the new functions.

src/sknnr/transformers/_tree_node_transformer.py
aazuspan: Yep, thanks for catching that!

src/sknnr/transformers/_tree_node_transformer.py
aazuspan: You definitely noticed something weird here that I'd missed. Looking closer at this, I would have expected two different `mypy` errors: an incompatible assignment to `X` when we re-assign it *and* an incompatible argument to `_set_tree_weights` when we pass the tuple. 

I spent some time trying to figure out exactly why we weren't seeing errors, and I think part of it was 1) `tuple[NDarray, NDArray]` is technically a subset of `ArrayLike`, so the re-assignment was valid. Weirdly though, `mypy` was eliminating the tuple type after re-assigning `X`, so it was just `NDArray` and matched `_set_tree_weights`. While that was actually correct in this case, I think it was just a fluke caused somehow by the re-assignment from `DataLike`. If I created a new variable `X_arr` instead of re-assigning to `X`, it *did* error due to the possible tuple type, as expected.

I still don't fully understand why that was happening, but I think the solution is to overload the return type for `_validate_data` to indicate that it depends on the `X` and `y` parameters, removing the ambiguity in the union return type. While I was making that change, I also went through and changed where we were previously re-assigning with `_validate_data` to instead set new variables. While the `DataLike` -> `NDArray` is a valid re-assignment, I wanted to err on the cautious side, given the weird behavior with the re-assignment before. Hopefully that's also a little more clear that `_validate_data` is doing an array conversion, but let me know if you think that's too heavy handed and we should revert the change.

> (I also noticed that we have a TODO on _validate_data to rework it when scikit-learn >= 1.6.0. I can't remember when we did that, but that's what shows in our `pyproject.toml' file.)

It looks like we only pin the test environment to `>=1.6.0`, while the runtime dependencies are still on `>=1.2.0`. I'm not sure what the right schedule is for pinning a dependency - [`1.6.0`](https://scikit-learn.org/stable/whats_new/v1.6.html) is almost 1.5 years old now, so maybe we should start considering that?

src/sknnr/_base.py
aazuspan: The more specific types are definitely nicer! 

I tried to specify dtypes in [61f7cb0](https://github.com/lemma-osu/sknnr/pull/112/commits/61f7cb0ba572a3c0d7496af72788de5b982eb61c) and [e09ddbd](https://github.com/lemma-osu/sknnr/pull/112/commits/e09ddbd19d947a5cbe76ad1f740d21bde129b301), but please double check that those look correct and make sense. In particular, I'm not as confident in the more ambiguous dtypes like `TreeNodeTransformer._validate_and_promote_targets` where the arrays have been promoted. I set those as `NDArray[np.object_ | np.number]` since I *think* they'll be either a numeric Numpy array or would have been promoted to `object` if they were categorical, but let me know if you can think of a better solution. 

For now, I did leave `_validate_data` without data types, since those theoretically work on *any* array.

src/sknnr/transformers/_base.py
grovduck: Thanks for finding that reference.  Sounds like we're good leaving `-> None` off on all our constructors.  Looks good to me.

src/sknnr/transformers/_tree_node_transformer.py
grovduck: > I would have expected two different `mypy` errors: an incompatible assignment to `X`

Ah, I didn't catch this first inconsistency, but I do understand what you're saying here.

> I think the solution is to overload the return type for `_validate_data` to indicate that it depends on the `X` and `y` parameters, removing the ambiguity in the union return type.

Ah, more overloads 😉.  I've looked at the overloads on `_validate_data` and they look good to me.

> While I was making that change, I also went through and changed where we were previously re-assigning with `_validate_data` to instead set new variables. While the `DataLike` -> `NDArray` is a valid re-assignment, I wanted to err on the cautious side, given the weird behavior with the re-assignment before. Hopefully that's also a little more clear that `_validate_data` is doing an array conversion, but let me know if you think that's too heavy handed and we should revert the change.

I like this change.  Very clear that `X` is meant to differ from `X_arr`.

> It looks like we only pin the test environment to `>=1.6.0`, while the runtime dependencies are still on `>=1.2.0`.

Sorry, you're absolutely right.  

> I'm not sure what the right schedule is for pinning a dependency - [`1.6.0`](https://scikit-learn.org/stable/whats_new/v1.6.html) is almost 1.5 years old now, so maybe we should start considering that?

If you want, I'm OK punting on this decision for now.  Obviously not related to this PR.  Let me know if you'd like me to open a new issue to track this?

src/sknnr/transformers/_tree_node_transformer.py
grovduck: Sorry for beating a dead horse, but how did you decide to type both `regressor_cls` and `classifier_cls` as `type[BaseEstimator]`?   My naive thinking would have to typed these much more rigidly, e. g. `regressor_cls` as `type[RandomForestRegressor] | type[GradientBoostingRegressor]`.  I'm mostly curious about the tradeoffs of each approach?

aazuspan: Conceptually I was thinking that using those specific types was leaking information from the subclasses into the superclass, since those parameters aren't *inherently* limited to those types and might include other estimators in the future. But since the annotations are largely for static analysis, this is probably a case where practicality beats purity, and it's better to be specific about what's *currently* supported.

Let me know what you think, but for now I grouped those estimators into `TreeRegressor` and `TreeClassifier` aliases to simplify expanding them in the future, and updated the parameter annotations to point to those.

aazuspan:
Thanks for slogging through 600 lines of type annotations, @grovduck. Still a couple loose ends to resolve, but I realized I hadn't responded to your review yet.

> The one thing that gives me pause here is the support for sparse matrices in MatrixLike? For TransformedKNeighborsRegressor, we explicitly have tags.input_tags.sparse = False. Maybe I'm confused on whether that would cause an issue? If we typed X as MatrixLike and provided a scipy.sparse.spmatrix, then wouldn't any of our estimators derived from TransformedKNeighborsRegressor fail to fit?

Yeah, good point - we definitely want to exclude the sparse type, so I guess the current definition for `DataLike` is probably accurate.

> And potentially can speed-up runtimes? I think I've read that from you.

Yeah, at least in theory the startup should be a little faster by skipping the imports at runtime.

src/sknnr/transformers/_tree_node_transformer.py
aazuspan: > If you want, I'm OK punting on this decision for now. Obviously not related to this PR. Let me know if you'd like me to open a new issue to track this?

+1 for punting. Yeah, if you want to create that issue, that'd be great. Getting rid of the compatibility shim (and hopefully the overloads with it) would be nice.

src/sknnr/transformers/_tree_node_transformer.py
grovduck: That makes good sense and I agree that the way to handle the typing is with the `TreeRegressor` and `TreeClassifier` aliases.  I didn't really explain myself too well, but I think the only estimators in the future would also have to be an estimator that emitted (forests of) node IDs, so I was concerned that just typing as `BaseEstimator` would be too permissive.  (We don't want folks putting yucky `LinearRegression` estimators in there after all!).

src/sknnr/_base.py
grovduck: Yes, I think those look good.  A few notes:

- The `CCA` and `CCorA` classes are still not well-typed, but I think using `np.float64` for `transform` and `fit_transform` is appropriate.
- Can `RawKNNRegressor.kneighbors` also be a return type of `NDArray[np.int64] | tuple[NDArray[np.float64], NDArray[np.int64]]` (just like `TransformedKNeighborsRegressor.kneighbors`)?
- Ticky-tack, but can we also type `test_estimators.py::X_y_yfit` as `tuple[NDArray[np.float64], NDArray[np.float64], NDArray[np.float64]]`?
- I'm pretty sure you're correct about the return type of `NDArray[np.object_ | np.number]` from `_validate_and_promote_targets`.  Categoricals will get turned into numeric, but strings stay as strings.  Based on the same logic as renaming `X` to `X_arr` from `_validate_data`, do you think it's worth changing `y` to `y_arr` on line 145 of `_tree_node_transformer.py` to make clear that it's been modified?
- I've noticed that you've set `X: NDArray[np.object_ | np.number]` on all `_set_tree_weights` signatures.  But, when used, isn't `X` only coming from the return of `_validate_data` which is now typed as a generic `NDArray`?  Wouldn't you want `X` typed the same way on `_set_tree_weights` (and `train_improvement`)?

Sorry, I hope those notes are easy to decipher.

grovduck:
> Thanks for slogging through 600 lines of type annotations

No, thank _you_.  I'm sure the actual typing is much more painful than the review.  I think I've responded to everything here, but let me know if I haven't.  I have it on my TODO list to create an issue about pinning scikit-learn to a later version.

src/sknnr/transformers/_tree_node_transformer.py
aazuspan: Yeah, good point! Narrowing it down was definitely the right call.

src/sknnr/_base.py
aazuspan: Good calls all around! I implemented all of those changes in [d6802ae](https://github.com/lemma-osu/sknnr/pull/112/commits/d6802aeda49ef75588d80dd128da9827f648a906) and [448b153](https://github.com/lemma-osu/sknnr/pull/112/commits/448b153ef5ec3414601af2e9be3eafff8094bfa4).

grovduck (approved):
Excellent!  Time to call this one done!

PR #119, opened 2026-05-07
Pin `scikit-learn` runtime version to `>=1.6.0`

### LINKED ISSUE #118: Update `scikit-learn` runtime version to `>=1.6.0`
As of e800255, the `scikit-learn` runtime dependency is pinned to `scikit-learn>=1.2.0`.  For the hatch `test` environment, the dependency is pinned to `scikit-learn>=1.6.0` based on changes to the estimator tag system in `scikit-learn` as detailed in #82 and #84.  We have also been using a compatibility shim (`_base._validate_data`) to handle changes in `scikit-learn` where they moved `_validate_data` from a method of `BaseEstimator` to a public utility function.

As of May 2026, `scikit-learn 1.6.0` is about 1.5 years old (release date December 9, 2024).  Given that we know that our tests are passing at `>=1.6.0`, it's probably safe to pin our runtime dependency to that as well.  This would allow us to remove the `_validate_data` shim and removing the recent overloads introduced in #112.    

A very quick browse of the [release notes](https://scikit-learn.org/stable/whats_new/v1.6.html#release-notes-1-6) note that `1.6.0` was a major release, but I didn't see anything problematic.

grovduck:
This is a simple update to pin the runtime version of `scikit-learn` to `>=1.6.0`, which allows removal of the existing `sknnr._base._validate_data` function.

Because the `scikit-learn` tag system was updated at `1.6.0` and the `sknnr` estimators use the `__sklearn_tags__` method to expose these tags, users at `scikit-learn<1.6.0` would not receive the full set of tag information, e.g. at `<1.6.0`
```python
# /// script
# dependencies = [
#   "scikit-learn<1.6.0",
#   "sknnr",
# ]
# ///

from sknnr import GNNRegressor

est = GNNRegressor()
print(est._more_tags())

# {'pairwise': False}
```
whereas at `>=1.6.0`
```python
# /// script
# dependencies = [
#   "scikit-learn>=1.6.0",
#   "sknnr",
# ]
# ///

from sknnr import GNNRegressor

est = GNNRegressor()
print(est.__sklearn_tags__())

# Tags(estimator_type='regressor', target_tags=TargetTags(required=True, one_d_labels=False, two_d_labels=False, positive_only=False, multi_output=True, single_output=True), transformer_tags=None, classifier_tags=None, regressor_tags=RegressorTags(poor_score=False), array_api_support=False, no_validation=False, non_deterministic=False, requires_fit=True, _skip_test=False, input_tags=InputTags(one_d_array=False, two_d_array=True, three_d_array=False, sparse=False, categorical=False, string=False, dict=False, positive_only=False, allow_nan=False, pairwise=False))
```

@aazuspan, I wasn't sure if there was more to do on this beside the update to `1.6.0` and the removal of `_validate_data`.  Maybe a bit cavalier, so let me know if I should be tackling other items in this PR.

Closes #118.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,6 @@ dependencies = ["pre-commit"]
 
 [tool.hatch.envs.test]
 dependencies = [
-    "scikit-learn>=1.6.0",
     "pytest",
     "pytest-cov",
     "pytest-regressions",
@@ -51,11 +50,7 @@ template = "test"
 python = ["3.11", "3.12", "3.13", "3.14"]
 
 [tool.hatch.envs.docs]
-dependencies = [
-    "mkdocs<2.0.0", 
-    "mkdocs-material", 
-    "mkdocstrings[python]"
-]
+dependencies = ["mkdocs<2.0.0", "mkdocs-material", "mkdocstrings[python]"]
 
 [tool.hatch.envs.docs.scripts]
 serve = "mkdocs serve --config-file docs/mkdocs.yml --livereload --watch ./README.md"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ authors = [
     { name = "Matt Gregory", email = "matt.gregory@oregonstate.edu" },
     { name = "Aaron Zuspan", email = "aaron.zuspan@oregonstate.edu" },
 ]
-dependencies = ["numpy", "scikit-learn>=1.2", "scipy"]
+dependencies = ["numpy", "scikit-learn>=1.6.0", "scipy"]
 
 [project.urls]
 Homepage = "https://github.com/lemma-osu/sknnr"
--- a/docs/pages/installation.md
+++ b/docs/pages/installation.md
@@ -13,6 +13,6 @@ pip install sknnr --pre
 ## Dependencies
 
 - Python >= 3.11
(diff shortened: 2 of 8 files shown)


pyproject.toml
```diff
@@ -12,7 +12,7 @@ authors = [
     { name = "Matt Gregory", email = "matt.gregory@oregonstate.edu" },
     { name = "Aaron Zuspan", email = "aaron.zuspan@oregonstate.edu" },
 ]
-dependencies = ["numpy", "scikit-learn>=1.2", "scipy"]
+dependencies = ["numpy", "scikit-learn>=1.6.0", "scipy"]
```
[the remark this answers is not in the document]
grovduck: Fixed.  (Completely missed that one.)

pyproject.toml
aazuspan: This should be removable now that it matches the package dependencies.

aazuspan (changes requested):
One very minor nit, but otherwise LGTM! I love a PR that can remove more code than it adds 🎉

pyproject.toml
grovduck: Yes, of course!  Good catch.

aazuspan (approved):
LGTM!

